PR #1, opened 2020-03-25
CocoaPods

i-schuetz:


i-schuetz:
Same as https://github.com/Co-Epi/app-ios/pull/2 without .ipa

PR #2, opened 2020-03-25
App center, CocoaPods

i-schuetz:


i-schuetz:
Same as #1  without .ipa

PR #3, opened 2020-03-25
Abstract away CentralManager to hopefully ease uni testing

johnsonh:
The goal here is to remove references to `CoreBluetooth` in our business logic (`Central.swift`), which should make it easier to do unit testing. If we care about unit testing, some solution is necessary here, particularly because `init()` is unavailable for a lot of the frequently used CB types (`CBPeripheral`, `CBService`, `CBCharacteristic`).

In this PR, I abstract away `CBCentralManager`, by creating a `CentralManager` protocol for our business logic to use. Both the real `CBCentralManager` and a future mock would conform. Plus a few random notes/questions.

On the delegate side, I create a delegate protocol which (would eventually) not reference `CoreBluetooth` at all for business logic. The implementation uses a dumb shim around the delegate to make it easier to use. In the implementation it should be safe to `as! CBCentralManager` because the implementation should only be used when the app is really running (vs testing).

I think this marginally improves testability of `CentralManager` related code. The problem is that to fully pull out all `CoreBluetooth` references gets pretty hairy, because there's a bunch of `CB` types - `CBPeripherals`, `CBServices`, `CBCharacteristics`, etc, that reference each other, are passed into each others functions, etc. 

I'm probably most of the way to there: https://github.com/Co-Epi/app-ios/compare/master...johnsonh:BLE-abstract?expand=1. Most of the types are concrete, rather than protocols, so I wrap them in business objects that we can define. I've successfully gotten rid of the `CoreBluetooth` dependency in `Central`, which I renamed to `BluetoothManager`. 

However, things got fairly hairy, I've introduced a lot of protocols and types, and it's probably riddled with bugs that I'd need to spend time ironing out. I'm happy to do so if we think it's worthwhile. Due to `init()` not being available for a lot of the CB types, I think mocking/testing this code will require some non trivial solution. Alternatively to fully abstracting away CB types, I found 2 libraries for bluetooth helper types:
- https://github.com/Rightpoint/RZBluetooth
- https://github.com/NordicSemiconductor/IOS-CoreBluetooth-Mock

Not sure how reliable they are.

### DIFF
--- a/CoEpi.xcodeproj/project.pbxproj
+++ b/CoEpi.xcodeproj/project.pbxproj
@@ -23,12 +23,13 @@
 		84BF973024281082001ADD6B /* DataRepresentable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 84BF972A24281082001ADD6B /* DataRepresentable.swift */; };
 		84BF973124281082001ADD6B /* Types+DataRepresentable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 84BF972B24281082001ADD6B /* Types+DataRepresentable.swift */; };
 		84FA00F22427E13000318104 /* Dependencies.swift in Sources */ = {isa = PBXBuildFile; fileRef = 84FA00F12427E13000318104 /* Dependencies.swift */; };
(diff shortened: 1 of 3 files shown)


i-schuetz:
Do we still need this?

PR #9, opened 2020-03-26
Configure Realm

i-schuetz:


CoEpi/domain/model/Contact.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
+struct Contact {
+    let cen: String
```
johnsonh: Should we link to the docs on CEN generation?

scottleibrand:
Ready to merge?

PR #18, opened 2020-03-27
Onboarding

rase-rocks:
Merging the Onboarding UI.

Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-27 at 15 59 41

Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-27 at 15 59 46


i-schuetz:
Preliminary comment: The formatting is very different to what we use. We should be consistent. Would it be possible to switch to our (regular/common) formatting style?

And then submit to review again. Otherwise it will be difficult to see changes.

rase-rocks:
I think your referring to 'aligned colons and ='. I'll strip out the alignment and resubmit

rase-rocks:
Reformatted to remove alignment.

CoEpi/ui/onboarding/AttributedString.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    init(baseFont: UIFont) {
+        
+        self.baseFont = baseFont
+        
+        let pointSize = baseFont.pointSize
+        let fontDescriptor = baseFont.fontDescriptor.withSymbolicTraits(.traitBold)
+            ?? UIFont.boldSystemFont(ofSize: pointSize).fontDescriptor
+        
+        let boldFont = UIFont(descriptor: fontDescriptor, size: pointSize)
```
i-schuetz: You can put this in an extension:

```swift
extension UIFont {

    func toBold() -> UIFont {
        let boldFontDescriptor = fontDescriptor.withSymbolicTraits(.traitBold)
            ?? UIFont.boldSystemFont(ofSize: pointSize).fontDescriptor
        return UIFont(descriptor: boldFontDescriptor, size: pointSize)
    }
}
```

CoEpi/ui/onboarding/AttributedString.swift
```diff
@@ -0,0 +1,71 @@
...
+        let attributedString = NSAttributedString(string: "\(value) ", attributes: attributes)
+        storage.append(attributedString)
+    }
+    
+    func bold(_ value: String) -> AttributableString {
+        
+        append(value, attributes: boldAttributes)
+        
+        return self
+
```
i-schuetz: I'd use blank lines only to separate block semantically. Seem unnecessary in small methods.

CoEpi/ui/onboarding/AttributedString.swift
```diff
@@ -0,0 +1,71 @@
...
+        append(value, attributes: normalAttributes)
+        
+        return self
+        
+    }
+    
+    func finish() -> NSAttributedString {
+        
+        let range = NSRange(location: 0, length: storage.length - 1)
+        return storage.attributedSubstring(from: range)
```
i-schuetz: Is this a defensive copy (otherwise it seems fine to just return `storage`)?

CoEpi/ui/onboarding/AttributedString.swift
```diff
@@ -0,0 +1,71 @@
...
+        ]
+        
+        self.normalAttributes = [
+            .font: baseFont
+        ]
+        
+    }
+    
+    func append(_ value: String, attributes: FontAttributes) {
+        let attributedString = NSAttributedString(string: "\(value) ", attributes: attributes)
```
i-schuetz: I'd expect this function to just append `value` and not prepend a space to it. Maybe create a new function `space` that inserts the space and call it explicitly when building the string.

CoEpi/ui/onboarding/Constants.swift
```diff
@@ -0,0 +1,14 @@
+import UIKit
+
+struct Constants {
+    
+    struct color {
```
i-schuetz: We probably should move this to a more central place, called `Theme` or similar. I imagine that these colors can be used in other places of the app.

CoEpi/ui/onboarding/AttributedString.swift
```diff
@@ -0,0 +1,71 @@
...
+        
+    }
+    
+}
+
+extension AttributableString {
+    
+    static func avenir(size: CGFloat) -> UIFont {
+        
+        return UIFont(name: "AvenirNext-Regular", size: size)!
```
i-schuetz: No need for `return` keyword here

CoEpi/ui/onboarding/AttributedString.swift
```diff
@@ -0,0 +1,71 @@
...
+        
+        self.baseFont = baseFont
+        
+        let pointSize = baseFont.pointSize
+        let fontDescriptor = baseFont.fontDescriptor.withSymbolicTraits(.traitBold)
+            ?? UIFont.boldSystemFont(ofSize: pointSize).fontDescriptor
+        
+        let boldFont = UIFont(descriptor: fontDescriptor, size: pointSize)
+        
+        self.boldAttributes = [
```
i-schuetz: Don't use `self` if not required.

CoEpi/ui/onboarding/GradientView.swift
```diff
@@ -0,0 +1,42 @@
...
+        if horizontalMode {
+            gradientLayer.startPoint = diagonalMode ? .init(x: 1, y: 0) : .init(x: 0, y: 0.5)
+            gradientLayer.endPoint = diagonalMode ? .init(x: 0, y: 1) : .init(x: 1, y: 0.5)
+        } else {
+            gradientLayer.startPoint = diagonalMode ? .init(x: 0, y: 0) : .init(x: 0.5, y: 0)
+            gradientLayer.endPoint = diagonalMode ? .init(x: 1, y: 1) : .init(x: 0.5, y: 1)
+        }
+    }
+    
+    func updateLocations() {
```
i-schuetz: All these functions can be private.

CoEpi/ui/onboarding/OnboardingButton.swift
```diff
@@ -0,0 +1,20 @@
...
+        super.awakeFromNib()
+        
+        backgroundColor = Constants.color.coepiPrimaryButton
+        layer.cornerRadius = 5
+        contentEdgeInsets = UIEdgeInsets(top: 20,
+                                         left: 40,
+                                         bottom: 20,
+                                         right: 40)
+        
+        setTitleColor(.white, for: .normal)
```
i-schuetz: Can't all these attributes be set in the xib?
See https://stackoverflow.com/a/15839914/930450 for corner radius

CoEpi/ui/onboarding/OnboardingViewController.swift
```diff
@@ -12,8 +20,91 @@ class OnboardingViewController: UIViewController {
...
+    
+    func setupGradient() {
+        guard let view = view as? GradientView else { return }
+        view.startColor = Constants.color.coepiBlue
+        view.endColor = Constants.color.coepiAqua
+        view.endLocation = 0.4
+    }
+    
+    func setupButtons() {
+        privacyButton?.setAttributedTitle(privacyButtonTitle, for: .normal)
```
i-schuetz: If you declare the outlets as implicitly wrapped optionals, you don't need to use `?` here.

CoEpi/ui/onboarding/OnboardingViewController.swift
```diff
@@ -12,8 +20,91 @@ class OnboardingViewController: UIViewController {
...
+            .normal(Strings.onboarding.about)
+            .finish()
+        
+        let tagLineTitle = AttributableString(baseFont: fontLarge)
+            .normal(Strings.onboarding.collaborate + ".")
+            .normal(Strings.onboarding.inform + ".")
+            .normal(Strings.onboarding.protect + ".")
+            .finish()
+        
+        trackLabel.attributedText = trackTitle
```
i-schuetz: Join declaration and assignment?

```swift
trackLabel.attributedText = AttributableString(baseFont: font)
...
```

CoEpi/ui/onboarding/OnboardingViewController.swift
```diff
@@ -12,8 +20,91 @@ class OnboardingViewController: UIViewController {
...
+        
+    }
+    
     @IBAction func onCloseClick(_ sender: UIButton) {
         viewModel.onCloseClick()
     }
+    
+}
+
+extension OnboardingViewController {
```
i-schuetz: Can be private

CoEpi/ui/onboarding/Strings.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+struct Strings {
+    
+    struct onboarding {
```
i-schuetz: We should use `SwiftGen` to generate these keys. What do you think about removing localization from this PR and we implement it properly in a future version (considering also that the bold functionality isn't localizable)?

CoEpi/ui/onboarding/AttributedString.swift
rase-rocks: Returning the substring is to remove the trailing space that will be added as above. Couple of ways I see round this, build the string by pushing to the end of an array of `NSAttributedString` instead and doing join in the final method. But in the final commit this class will be likely removed. I will make the change and keep the code just in case it becomes useful later. I see the major problem with it being translating the emphasis (bold etc) into right to left languages. See the next commit for the altered solution

CoEpi/ui/onboarding/AttributedString.swift
rase-rocks: As with the above I'm unhappy with this class as a whole. Fixing fonts with magic strings in forced unwrapped optionals. Hmm.

CoEpi/ui/onboarding/Constants.swift
rase-rocks: Good idea. I will give them better names also

CoEpi/ui/onboarding/OnboardingButton.swift
rase-rocks: They can, however this button is going to get reused in future. So, it removed the need for manually updating the various nibs across the project. But I'm happy to remove it if preferred

CoEpi/ui/onboarding/OnboardingViewController.swift
rase-rocks: The unwrapping was added to prevent a crash in the Unit Test target, I'm not familiar with the way the view controllers are setup / how to get a fully initialised one for use in tests. So until I work that out optional chaining prevents the crash.

CoEpi/ui/onboarding/Strings.swift
rase-rocks: Done. There is now a struct containing the new markup strings, just to keep it out of the view controller. Localization has been removed.

rase-rocks:
Since the original PR the ability to auto-merge has passed with the above conflicts. The Xcode project conflict creates an issue

CoEpi/ui/onboarding/AttributedString.swift
i-schuetz: Another reason to not add those spaces in `append` 😉

CoEpi/ui/onboarding/OnboardingButton.swift
i-schuetz: It's a fair point, but styling isn't a good reason for subclassing. We should create reusable styles. I'll add a TODO for now.

i-schuetz: It's a fair point, but subclassing isn't a good fit for styling, in particular because it composes poorly. We should use either reusable style objects/functions or protocol extensions. I'll add a TODO for now.

CoEpi/ui/onboarding/OnboardingViewController.swift
i-schuetz: I ran the unit tests and it didn't crash. Some points:
- We shouldn't be unit testing view controllers anyway IMO. Unit Tests make more sense for business logic.
- If it crashes it would be interesting to figure out anyway. Can you check again?
- In case we really needed to use `?`, the variables should be declared as optionals, to be consistent.
- I'll remove the `?` since it's not crashing and folks want a build. Let's come back to it if it crashes for you.

PR #23, opened 2020-03-28
Roboto font added

duskoo:


CoEpi/ui/healthquiz/QuestionCell.swift
```diff
@@ -14,6 +15,17 @@ class QuestionCell: UITableViewCell {
             this.onChecked?(q, checked)
         }
         accessoryView = checkBox
+        guard let customFont = UIFont(name: "Roboto-Regular", size: 18) else {
```
i-schuetz: We should put the fonts in a central place (`Fonts` or however you want to name it), to not have to repeat this code and cache them.

CoEpi/ui/home/HomeViewController.swift
```diff
@@ -31,6 +37,8 @@ class HomeViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         
+        //==Family: Roboto Font names: ["Roboto-Regular", "Roboto-Light", "Roboto-Bold"]
```
i-schuetz: Delete comment if not needed.

rase-rocks:
Here is an example of #27 in action. If this could be resolved it could reduce the amount of fiddling required for every commit

CoEpi/ui/healthquiz/QuestionCell.swift
duskoo: Added CoEpi/ui/resources/fonts/Fonts.swift

duskoo:
Closing this PR. Please take a look at #29

PR #29, opened 2020-03-29
Roboto font rebased

duskoo:
Rebased PR #23


CoEpi/ui/home/HomeViewController.swift
```diff
@@ -17,6 +17,12 @@ class HomeViewController: UIViewController {
         self.viewModel = viewModel
         super.init(nibName: String(describing: Self.self), bundle: nil)
         self.title = self.viewModel.title
+
+        let attrs = [
+            NSAttributedString.Key.font: Fonts.robotoRegular
+        ]
+
+        UINavigationBar.appearance().titleTextAttributes = attrs
```
i-schuetz: Join declaration and assignment.

CoEpi/ui/healthquiz/QuestionCell.swift
```diff
@@ -14,6 +15,10 @@ class QuestionCell: UITableViewCell {
             this.onChecked?(q, checked)
         }
         accessoryView = checkBox
+
+        textLabel?.font = UIFontMetrics.default.scaledFont(for: Fonts.robotoRegular)
```
i-schuetz: This is still creating a new font per cell (unless the system caches it). Put it in `Fonts`.

CoEpi/ui/resources/fonts/Fonts.swift
```diff
@@ -0,0 +1,47 @@
+//
+//  Fonts.swift
+//
+//  Created by Dusko Ojdanic on 3/29/20.
+//  Copyright © 2020 coepi.org. Licensed under MIT License
+//
+// See LICENSE.txt for license information.
+//
```
i-schuetz: Remove header.

CoEpi/ui/healthquiz/QuestionCell.swift
duskoo: @i-schuetz It seems this only sets a reference to the existing font object.
Screen Shot 2020-03-30 at 3 10 47 AM

CoEpi/ui/resources/fonts/Fonts.swift
```diff
@@ -0,0 +1,39 @@
...
+enum FontNames : String {
+    case robotoRegular = "Roboto-Regular"
+    case robotoLight = "Roboto-Light"
+    case robotoBold = "Roboto-Bold"
+}
+
+class Fonts {
+    static var robotoRegular : UIFont = { guard let font = UIFont(name: FontNames.robotoRegular.rawValue , size: 18)
+        else {
+            fatalError(fatalFontErrorMessage(font: .robotoRegular))
```
johnsonh: Should we fatal if a font isnt there?

i-schuetz: I think it's okay, since it's clearly a programmatic error.

At a later point we can default to preferred fonts and so on but this feels like overkill atm.

CoEpi/ui/healthquiz/QuestionCell.swift
i-schuetz: Well, this is not cell specific. But I realized that caching the scaled font would prevent it from reacting to accessibility changes (which the user may do while they're using the app). So this should be fine.

PR #30, opened 2020-03-30
iOS 12 compatibility

danamlewis:
Following various ([1](https://sarunw.com/posts/create-new-ios12-project-in-xcode11/), [2](https://stackoverflow.com/a/60294314)) bread crumbs, it looks like the following steps are needed to add iOS 12 compatibility:

TODO:

- [x] Target iOS 12 as deployment target in xcode 

(Un-did, is not needed) fixing AppDelegate to enable iOS 12

- [ ] unclear if needs to be done, adjusting `var window: UIWindow?` from the current ` internal var window: UIWindow?` Seems to work fine in initial local iOS12 build. 

Don't need to do (per [this rec](https://forums.developer.apple.com/thread/118484)) because these don't exist in our repo:

- Remove Scene Delegate
- Remove Application Scene Manifest from info.plist


CoEpi/AppDelegate.swift
```diff
@@ -11,7 +11,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     internal var window: UIWindow?
```
danamlewis: per the guide to making ios12 compatible - @i-schuetz does it matter if this is internal var vs just `var window: UIWindow?`?

CoEpi/AppDelegate.swift
```diff
@@ -11,7 +11,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
     
     private var rootWireframe: RootWireFrame?
-
+    
+    @available(iOS 13.0, *)
```
danamlewis: ditto a question on making it iOS12 compatible - it looks like this was needed for ScreenDelegate, which we're not calling - so I should pull this back out? @i-schuetz

CoEpi/AppDelegate.swift
johnsonh: It's internal by default. If you can make it private that'd be ideal but if not, leaving it as just `var window: UIWindow?` should be fine

CoEpi/AppDelegate.swift
johnsonh: I imagine it should be fine? As long as you test on iOS 12 and everything still works

danamlewis: k, thanks. All of the examples of making iOS12 compatible referenced it without it being internal so wasn't sure if our current use would snafu on iOS12.

CoEpi/AppDelegate.swift
danamlewis: k, thanks. All of the examples of making iOS12 compatible referenced it without it being internal so wasn't sure if our current use would snafu on iOS12. will let @i-schuetz make the call on whether to leave as-is (internal) or not, and confirm with iOS 12 testing later.

danamlewis:
tested locally and works on an iOS12 device. @scottleibrand review?

CoEpi/AppDelegate.swift
scottleibrand: When I load the debug screen on an iPhone 6s running iOS 12.4, there's no way to slide it down to get back to the home screen, and I have to kill the app to get out of the debug screen. Is this likely related, or something different?

scottleibrand: I suspect it's unrelated, so I'll open a separate issue.

PR #33, opened 2020-03-30
Hook up Alert acknowledge button + VM fix for health quiz

johnsonh:
On pressing acknowledge (based on this thread https://coepi.slack.com/archives/CUR0KQRNG/p1585473955094600)
- remove the alert from the UI
- remove the alert from repo so leaving/returning the UI maintains acknowledge state. would need to save acknowledged state on the backend to maintain across app deletes
- Also fixed VM for health quiz

alert-acknowledge-2

The acknowledge button doesn't work (its `IBAction` doesnt get called) with `AlertsViewController.xib` in develop. I need to fix it in a separate PR. In the meantime, the acknowledge button works in intermediate commits from the Alerts UI work, e.g. commit `98cad4c9d69641a8ea3b2fd1d394ece1fb0b1800` in this commit history (if anyone wants to test)


CoEpi/repo/AlertRepo.swift
```diff
@@ -13,4 +15,9 @@ class AlertRepoImpl: AlertRepo {
     func alerts() -> [Alert] {
         alertData
     }
+
+    func removeAlert(alert: Alert) {
```
i-schuetz: We should use rx in the repos too. So
- `alertData` becomes an observable (a subject, such that you can push data to it)
- In the view model you just have to map this observable to the outputs of the view model.
- When removing, you can either manipulate directly the value and push a new one or push a "delete event" that is combined with the alerts observable and pushes a new value to it. The later is relatively complicated. So it's ok if you implement the former.

CoEpi/ui/alerts/AlertsViewController.swift
```diff
@@ -24,12 +16,17 @@ class AlertsViewController: UIViewController {
         super.init(nibName: String(describing: Self.self), bundle: nil)
         
         title = "Alerts"
+        dataSource.onAcknowledged = acknowledge
```
i-schuetz: Why not pass the callback directly here?

CoEpi/ui/alerts/AlertsViewModel.swift
```diff
@@ -4,19 +4,21 @@ import RxSwift
 
 class AlertsViewModel {
     private let alertRepo: AlertRepo
+    private let alertsRelay: BehaviorRelay<[Alert]>
```
i-schuetz: If we add rx to the repo we don't need this anymore.

CoEpi/repo/AlertRepo.swift
johnsonh: - I think we'd want it to still be a Relay right? Relays are `A Subject except without the ability to call onComplete or onError`

But will make this change

johnsonh: Updated, not sure if I did this right. I just moved the `BehaviorRelay` into the alert repo and made it part of the interface, as is done for `Central` and `Peripheral`.

When you say mapping the observable, are you referring to making `Driver`s out of it? Or mapping it to a `AlertViewState` object from the domain `Alert` model?. Don't think I'm following the combining observables portion... I think it's my inexperience with Rx, I assume it's not just having the relay accept a new value, which pushes state to the driver?

CoEpi/repo/AlertRepo.swift
```diff
@@ -1,16 +1,31 @@
...
-    func alerts() -> [Alert]
+//    func alerts() -> [Alert]
+    var alerts: BehaviorRelay<[Alert]> { get }
+
+    func removeAlert(alert: Alert)
 }
 
 class AlertRepoImpl: AlertRepo {
-    let alertData: [Alert] = [
+    private(set) var alerts: BehaviorRelay<[Alert]>
```
i-schuetz: The relay itself doesn't have to be mutable. Just set the array directly here.

CoEpi/repo/AlertRepo.swift
i-schuetz: For now a Relay is fine. Though at some point this will come from an api or the local database and we'd need error handling (unless we model the error flow via a `Result` type. But it's probably better to stick to the default rx error handling).

What I mean with mapping is deriving new observables from this observable, which you're already doing in the view model.

CoEpi/repo/AlertRepo.swift
```diff
@@ -1,16 +1,31 @@
+import RxCocoa
+import RxSwift
+
 protocol AlertRepo {
-    func alerts() -> [Alert]
+//    func alerts() -> [Alert]
```
i-schuetz: Delete

CoEpi/ui/alerts/AlertsViewModel.swift
```diff
@@ -27,4 +26,12 @@ class AlertsViewModel {
         }
         return title + "s"
     }
+
+    public func acknowledge(alert: Alert) {
+        var newValue: [Alert] =  alertRepo.alerts.value
+        guard let idx = newValue.firstIndex(where: { $0.id == alert.id }) else { return }
+        newValue.remove(at: idx)
```
i-schuetz: Did you forget to remove this?

CoEpi/ui/healthquiz/HealthQuizViewModel.swift
```diff
@@ -5,29 +5,31 @@ class HealthQuizViewModel {
...
+    private let questionsRelay: BehaviorRelay<[Question]>
 
-    private var questions: [Question]
-    
     private(set) var rxQuestions: Driver<[Question]>
     
     init(container: DependencyContainer) {
         self.symptomRepo = try! container.resolve()
-        questions = symptomRepo.symptoms().map{ Question(symptom: $0) }
+        questionsRelay = BehaviorRelay(value: symptomRepo.symptoms()
```
i-schuetz: Since you're already working on this, you could move the `Relay` here as well to the repo. I didn't do it because we didn't need to edit the symptoms yet, but we will need it.

johnsonh: I'm not sure we want to do that. The symptom repo is the canonical source of all Symptoms, vs the behavior relay here is for Questions. For example, if a user selected a few answers in a quiz, and then did a quiz again, we wouldn't want the first quiz to impact the 2nd quiz (e.g. missing questions/symptoms, prechecked ones, etc.)

Questions drive view state, vs symptoms are the full set of possible symptoms, afaik.

i-schuetz: I didn't express myself well, I mean basically only to make the symptoms reactive. But I realize now that it's not trivial to wire this "rx-way" with the question inputs. You can let it for now like this. We will have to wire it in any case later, if/when we get the symptoms from a backend.

PR #34, opened 2020-03-30
Disable dark mode

i-schuetz:


danamlewis:
Noting that we'll support dark mode in future versions (card [here](https://github.com/orgs/Co-Epi/projects/1#card-35426232)), disabling for now so we can focus on 0.1 basics.

PR #38, opened 2020-03-31
Bridge improvements

duskoo:


CoEpi/ble/Central.swift
```diff
@@ -86,8 +100,12 @@ class Central: NSObject {
     }
     #endif
 
-    init(delegate: CentralDelegate) {
-        self.delegate = delegate
+//    init(delegate: CentralDelegate) {
```
johnsonh: Remove?

CoEpi/ble/Peripheral.swift
```diff
@@ -104,24 +113,25 @@ extension Peripheral: CBPeripheralManagerDelegate {
...
+
     private func addNewContactEvent(with identifier: String) {
         print("PERIPHERAL: addNewContactEvent called")
-        delegate?.onPeripheralContact(CEN(
-            identifier: identifier,
-            timestamp: Date(),
-            // TODO preference, from React Native
-            isPotentiallyInfectious: true
-        ))
+        delegate?.onPeripheralContact(CEN(CEN: identifier))
```
johnsonh: Remove?

duskoo: Idea is to keep both: rx relays and a delegate, for the time being (as suggested by @i-schuetz ).

CoEpi/ble/Central.swift
```diff
@@ -11,15 +11,21 @@ struct DetectedPeripheral {
...
+protocol CentralDelegate: class {
+    func onDiscovered(peripheral: CBPeripheral)
+    func onCentralContact(_ contact: CEN)
+}
 
-protocol Central {
+protocol CentralReactive {
     var discovery: PublishRelay<DetectedPeripheral> { get }
     var centralContactReceived: PublishRelay<ContactOld> { get }
+    var delegate: CentralDelegate? {get set}
```
i-schuetz: Why will be support both delegate and rx? This leads to a lot of confusion and is error prone. Let's stick to one or the other. From the last chats it seems we agreed on RX.

CoEpi/ble/Peripheral.swift
i-schuetz: I see now the cause of the confusion. This is not what I meant! I mean that if we have to work with delegates (because the covidwatch library most likely will not support RX), we can bridge those to RX by wrapping them in an adapter.

For now the plan of action is to just stick to one of the models. We agreed on RX. So delegate has to be removed.

i-schuetz:
Not important anymore, but it would have been better to create a branch of @rodneywitcher 's branch so we can see the deltas to that branch instead of the whole develop changeset.

CoEpi/ble/Peripheral.swift
i-schuetz: I see now the cause of the confusion. This is not what I meant! I mean that if we have to work with delegates (because the covidwatch library most likely will not support RX), we can bridge those to RX by wrapping them in an adapter.

For now the plan of action is to just stick to one of the models. We agreed on RX. So delegate has to be removed.

duskoo: Ok. Will make the change in #40

PR #40, opened 2020-03-31
Refactor CEN, CENKey, CENReport into Realm/Repo structure, change Peripheral interface/impl to be use case agnostic

johnsonh:
# Part 1, up until commit `fa84a6e`
Move all the files from the `/storage` folder into their proper places
- `DB*` Classes into `/realm/model`
- CEN, CENKey, CENReport into `/domain/model`
- All the logic that was in CEN, CENKey, CENReport into `/repo/realm` implementation
- Create repo protocols for CEN, CENKey, CENReport
- Move utils to `/repo`
- No logic changes (even the repo code, which could use some `map`, `filter`, etc. love

Doesn't compile because it doesn't set up the dependencies (yet)

# Part 2, mostly commit `ac1a011`
Refactor `Peripheral` to be agnostic of `CEN`, db/storage, etc. (`Central` is already nice and CEN/realm agnostic). 
- Create `PeripheralRequestHandler` protocol to that handles a received contact + associated dependencies (aka generate CEN and save to db). 
- Modify the public delegate/reactive protocols to pass `Data`. The next step would be to make `Peripheral` generic on `theirData` and `ourData` - in this case, I believe`ourData` is a `CEN` and `CENKey`, and I think we don't care about `theirData` (but other use cases might). 

Alternatively, we could just instantiate the repos inside of `Peripheral`, baking CEN and db into the bluetooth code, and it would be simple and done. Could be more prudent until we have a better understanding of the use cases of other apps (and is maybe better suited for the rust repo?)

# Testing

I haven't been able to test anything because I don't actually have an iOS device LOL. Would really appreciate help testing.

# Follow Up Work
- More robust Network layer https://github.com/Co-Epi/app-ios/pull/25/#pullrequestreview-384039940
- Update `Cen` to be `ReceivedCen`, like in Android https://github.com/Co-Epi/app-ios/pull/40#discussion_r401482304
- Add API Models https://github.com/Co-Epi/app-ios/pull/40#discussion_r401482304
- Remove `Contact` related code (repos, models, etc.) https://github.com/Co-Epi/app-ios/pull/40#discussion_r401482304
- Make repo code reactive https://github.com/Co-Epi/app-ios/pull/40#discussion_r400912226


CoEpi/ble/ContactReceivedHandler.swift
```diff
@@ -0,0 +1,22 @@
...
+import Foundation
+
+class ContactReceivedHandler: PeripheralRequestHandler {
+
+    private let cenRepo: CENRepo
+    private let cenKeyRepo: CENKeyRepo
+
+    init(container: DependencyContainer) {
+        self.cenRepo = try! container.resolve()
+        self.cenKeyRepo = try! container.resolve()
```
i-schuetz: Inject specific dependencies. I saw we have this anti-pattern everywhere. I did this for the root wireframe exceptionally because it may need to resolve a lot of dependencies and it becomes impractical, though also discussable.

CoEpi/ble/Peripheral.swift
```diff
@@ -1,33 +1,43 @@
-import Foundation
 import CoreBluetooth
+import Dip
+import Foundation
 import os.log
 import RxRelay
 
+struct Contact2 {
```
i-schuetz: ?

CoEpi/ble/Peripheral.swift
```diff
@@ -1,33 +1,43 @@
...
-    func onPeripheralContact(_ contact: CEN)
+    func onPeripheralContact(_ contact: Contact2)
 }
 
 protocol PeripheralReactive {
     var peripheralState: BehaviorRelay<String> { get }
 
-    var peripheralContactSent: PublishRelay<ContactOld> { get }
+    var peripheralContactSent: PublishRelay<Contact2> { get }
     var delegate: PeripheralDelegate? {get set}
```
i-schuetz: As mentioned in other PR, let's get rid of this duplication

CoEpi/ble/Peripheral.swift
```diff
@@ -107,34 +117,26 @@ extension PeripheralImpl: CBPeripheralManagerDelegate {
...
         os_log("Peripheral manager did receive read request: %@", log: blePeripheralLog, request.description)
+
+        let respondClosure: (Data) -> () = { [weak self] (responseData) in
+            // do we need to capture request & peripheral?
+            let theirData: Data? = request.value
+            request.value = responseData
+            peripheral.respond(to: request, withResult: .success)
+
+            let contact: Contact2 = .init(theirData: theirData, ourData: responseData)
+            self?.addNewContactEvent(contact: contact)
```
i-schuetz: Do we have to do anything when someone reads our CEN?

CoEpi/ble/Peripheral.swift
duskoo: We already have Contact and ContactOld...

CoEpi/domain/model/CENKey.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+import CryptoKit
+import RealmSwift
+import Security
+
+struct CENKey : Codable {
+    var cenKey: String?
+    var timestamp: Int64 = Int64(Date().timeIntervalSince1970)
```
i-schuetz: Why mutable fields? And the inconsistency with how `CEN` is initialized?

CoEpi/ble/Peripheral.swift
```diff
@@ -1,33 +1,43 @@
...
+    func onReceivedRequest(request: Data?, respondClosure: (Data) -> ())
+}
 
 class PeripheralImpl: NSObject, PeripheralReactive {
     let peripheralState: BehaviorRelay<String> = BehaviorRelay(value: "unknown")
-    let peripheralContactSent: PublishRelay<ContactOld> = PublishRelay()
+    let peripheralContactSent: PublishRelay<Contact2> = PublishRelay()
     weak var delegate: PeripheralDelegate?
-
+    private let internalDelegate: PeripheralRequestHandler
```
duskoo: What is the point of adding another delegate? There is already var delegate: PeripheralDelegate

CoEpi/domain/model/CENReport.swift
```diff
@@ -0,0 +1,8 @@
+struct CENReport: Codable {
+    let CENReportID: String
+    let report: String
+    let reportMimeType: String
+    let reportTimestamp: Int64
+    let CENKeys: String
+    let isUser: Bool
```
i-schuetz: Is this old or very recent? Report seems to have only 3 fields? https://github.com/Co-Epi/coepi-backend-go

CoEpi/repo/CENKeyRepo.swift
```diff
@@ -0,0 +1,11 @@
+protocol CENKeyRepo {
```
i-schuetz: This should all be rx, but doesn't have to be in this PR

CoEpi/repo/Utils.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+let CENKeyLifetimeInSeconds: Int64 = 2*60 //TODO: revert back to 7*86400
+let CENLifetimeInSeconds: Int64 = 1*60 //TODO: revert back to 15*60
+
+func roundedTimestamp(ts : Int64) -> Int64 {
+    return Int64(ts / CENKeyLifetimeInSeconds)*CENKeyLifetimeInSeconds
```
i-schuetz: No need for `return`

CoEpi/ble/Peripheral.swift
```diff
@@ -1,33 +1,43 @@
-import Foundation
 import CoreBluetooth
+import Dip
+import Foundation
 import os.log
 import RxRelay
 
+struct Contact2 {
+    let theirData: Data?
```
duskoo: The logic for using Data type is... ?

CoEpi/repo/realm/RealmCENKeyRepo.swift
```diff
@@ -1,36 +1,39 @@
...
+import Foundation
 import RealmSwift
 import Security
 
-struct CENKey : Codable {
-    var timestamp: Int64 = Int64(Date().timeIntervalSince1970)
-    var cenKey: String?
-    
+class RealmCENKeyRepo: CENKeyRepo, RealmRepo {
+    var realm: Realm
```
i-schuetz: `let`

CoEpi/repo/realm/RealmCENKeyRepo.swift
```diff
@@ -1,36 +1,39 @@
...
             //generate base64string representation of key
             let cenKeyString = computeSymmetricKey()
             let cenKeyTimestamp = curTimestamp
-            
+
             //Create CENKey and insert/save to Realm
-            let newCENKey = CENKey(timestamp: cenKeyTimestamp, cenKey: cenKeyString)
-            newCENKey.insert()
+            let newCENKey = CENKey(cenKey: cenKeyString, timestamp: cenKeyTimestamp)
+            let _ = insert(key: newCENKey)
```
i-schuetz: Can remove `let` here

CoEpi/repo/realm/RealmCENKeyRepo.swift
```diff
@@ -1,36 +1,39 @@
...
-    static func generateAndStoreCENKey() -> CENKey {
+    var cenKeyTimestamp: Int64 = 0
+
+    init(realmProvider: RealmProvider) {
+        realm = realmProvider.realm
+    }
+
+    func generateAndStoreCENKey() -> CENKey {
         //Retrieve last cenKey and cenKeyTimestamp from CENKey
         let latestCENKey = getLatestCENKey()
```
i-schuetz: I know this is not your code and we're focusing on high level, but anyway, maybe remove `getLatestCENKey()` to the end of the function as it's used only here.

CoEpi/repo/realm/RealmCENKeyRepo.swift
```diff
@@ -1,36 +1,39 @@
...
-            
+
             //Create CENKey and insert/save to Realm
-            let newCENKey = CENKey(timestamp: cenKeyTimestamp, cenKey: cenKeyString)
-            newCENKey.insert()
+            let newCENKey = CENKey(cenKey: cenKeyString, timestamp: cenKeyTimestamp)
+            let _ = insert(key: newCENKey)
             return newCENKey
         } else {
             return latestCENKey!
```
i-schuetz: We have to handle these optionals properly, this seems unsafe.

CoEpi/repo/realm/model/RealmCENKey.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+import RealmSwift
+
+final class RealmCENKey : Object {
+    @objc dynamic var CENKey: String = ""
+    @objc dynamic var timestamp: Int64 = Int64(Date().timeIntervalSince1970)
+    
+    override static func primaryKey() -> String? {
+        return "timestamp"
```
duskoo: TImestamp is the primary key?

CoEpi/domain/model/CENKey.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+import CryptoKit
+import RealmSwift
+import Security
+
+struct CENKey : Codable {
+    var cenKey: String?
```
i-schuetz: This shouldn't be optional. If there's not key then this struct doesn't make sense.

CoEpi/repo/realm/RealmCENRepo.swift
```diff
@@ -2,54 +2,54 @@ import Foundation
...
+            return false
+        }
+    }
+
+    func loadAllCENRecords() -> [CEN]? {
+        let DBCENObject = realm.objects(RealmCEN.self).sorted(byKeyPath: "timestamp", ascending: false)
+        return DBCENObject.map { CEN(CEN: $0.CEN, timestamp: $0.timestamp) }
+    }
+
+    func generateCENData(CENKey : String) -> Data {
```
duskoo: generateCENData has nothing to do with Realm.

CoEpi/repo/realm/RealmCENKeyRepo.swift
```diff
@@ -43,45 +46,48 @@ struct CENKey : Codable {
...
                 }
             }
             return retrievedCENKeyList
         }
     }
-    
-    func insert() {
+
+    func insert(key: CENKey) -> Bool {
         let realm = try! Realm()
```
i-schuetz: Use `realm` in the repo's field.

CoEpi/repo/realm/RealmCENKeyRepo.swift
```diff
@@ -43,45 +46,48 @@ struct CENKey : Codable {
...
-        let sameObject = realm.objects(DBCENKey.self).filter("timestamp = %@", self.timestamp)
+        let sameObject = realm.objects(RealmCENKey.self).filter("timestamp = %@", key.timestamp)
         if sameObject.count > 0 {
             //Duplicate Entry: NOT inserting
+            return false
         } else {
-            let newCENKey = DBCENKey(_ts: self.timestamp, _cenKey: self.cenKey!)
+            let newCENKey = RealmCENKey(key)
             try! realm.write {
```
i-schuetz: Use repo's `write`

CoEpi/repo/realm/RealmCENRepo.swift
```diff
@@ -2,54 +2,54 @@ import Foundation
...
-struct CEN : Codable {
-    var CEN: String
-    var timestamp: Int64 = Int64(Date().timeIntervalSince1970)
-    var latitude: Double? = nil
-    var longitude: Double? = nil
-    
-    static func generateCENData(CENKey : String)  -> Data {
+class RealmCENRepo: CENRepo, RealmRepo {
+
+    var realm: Realm
```
i-schuetz: `let`

CoEpi/repo/realm/RealmCENRepo.swift
```diff
@@ -2,54 +2,54 @@ import Foundation
...
+
+    init(realmProvider: RealmProvider) {
+        realm = realmProvider.realm
+    }
+
+    func insert(cen: CEN) -> Bool {
+        let DBCENObject = realm.objects(RealmCEN.self).filter("CEN = %@", cen.CEN)
+        if DBCENObject.count == 0 {
+            let newCEN = RealmCEN(cen)
+            try! realm.write {
```
i-schuetz: Use repo's `write` (and in similar places below, same with `realm`)

CoEpi/repo/realm/model/RealmCENReport.swift
```diff
@@ -19,21 +19,17 @@ final class DBCENReport : Object {
...
-    
-    required init() {
-        //fatalError("init() has not been implemented")
-        self.CENReportID = NSUUID().uuidString
-        self.reportTimestamp = 0
+
+    convenience init(_ report: CENReport) {
+        self.init()
+
+        self.CENReportID = report.CENReportID
```
i-schuetz: Let's use `self` only when the compiler requires it, as a convention.

i-schuetz:
Looking good! Let's not be so strict for now with the target PR's code, since we have time pressure, it's a lot and it will likely be replaced with the Rust lib. See what makes sense to improve and request review again.

CoEpi/ble/Peripheral.swift
johnsonh: The point is to be able to pass the CEN data to consumers of the BT code, but not have the BT code know that it's CEN data/know what it is. So the BT code can just pass it out as Data without knowing what data it is. That's what this comment was about:
`next step would be to make Peripheral generic on theirData and ourData`

So the consumer can know that it's actually CENs but the Peripheral code just sees it as data

CoEpi/ble/Peripheral.swift
johnsonh: The point is to be able to pass the CEN data to consumers of the BT code, but not have the BT code know that it's CEN data/know what it is. So the BT code can just pass it out as Data without knowing what data it is. That's what this comment was about:
next step would be to make Peripheral generic on theirData and ourData

So the consumer can know that it's actually CENs but the Peripheral code just sees it as data

CoEpi/ble/ContactReceivedHandler.swift
johnsonh: Ah of course, this is obvious I just didn't think about it. Done

CoEpi/domain/model/CENKey.swift
johnsonh: Made lets. What do you mean by inconsistency with how CEN is initialized?

CoEpi/ble/Peripheral.swift
johnsonh: Renamed to `BTContact`, or are you suggesting we shouldn't have this at all? See https://github.com/Co-Epi/app-ios/pull/40#discussion_r400971723,

CoEpi/ble/Peripheral.swift
johnsonh: That delegate handles when the peripheral is connected to, aka when we generate keys and save to DB. The point is so that no CEN/DB code lives in the peripheral - the peripheral just calls some delegate.

CoEpi/repo/realm/model/RealmCENKey.swift
johnsonh: yikes, fixing

CoEpi/repo/realm/RealmCENRepo.swift
johnsonh: Agreed, moved it to `ContactReceivedHandler`

CoEpi/ble/Peripheral.swift
johnsonh: What do you mean? I mean we need to callback the consumer so that we can do stuff like the `DebugViewController`, monitoring what peripherals we've connected with

CoEpi/domain/model/CENReport.swift
johnsonh: I guess we should develop against https://github.com/Co-Epi/data-models/pull/4/files#diff-5fb097d2df59c1b9d327693c1242ae4dR101?  But on the other hand it sounds like reports will change over time https://coepi.slack.com/archives/CUQGSQQ3A/p1585659605266100?thread_ts=1585638001.264000&cid=CUQGSQQ3A

Regardless I think we can remove the ones that aren't in the API definition

CoEpi/repo/CENKeyRepo.swift
johnsonh: How would we make this reactive?

CoEpi/repo/realm/RealmCENKeyRepo.swift
johnsonh: Not really sure what the default behavior should be...

CoEpi/ble/Peripheral.swift
i-schuetz: I think that this currently is distracting. At this moment we don't plan to do anything with this event. Let's tailor the interface to the spec. Add a log if needed.

CoEpi/domain/model/CENKey.swift
i-schuetz: Aside of `var`, that you use `init` there.

CoEpi/repo/CENKeyRepo.swift
i-schuetz: For "setters" we would return `Completable` and for "getters" ideally we'd maintain observable fields in the repo, such that potentially multiple observers (e.g. different screens) can subscribe to them.

CoEpi/ble/Peripheral.swift
i-schuetz: At the moment all what we get from BT (correct me if I'm wrong) is a byte array, which we did read from their characteristic or advertisement data.

Either return only the byte array and convert to a CEN type (this would be just a wrapper for a byte array/data) or couple BT directly with CEN.
I don't find the later so bad, given that the covadwitch plans to do the same and it helps us a little with readability and safety. The way the BT is used is tailored around contact tracing, so this coupling makes sense.

CoEpi/repo/realm/RealmCENKeyRepo.swift
i-schuetz: Yeah let's keep the code "as is" for now.

CoEpi/repo/CENKeyRepo.swift
i-schuetz: On the other side, keeping the api synchronous isn't necessarily bad, as it can be also easily composed with other synchronous apis. The threading / rx would be added "outside".

CoEpi/ble/Peripheral.swift
i-schuetz: Another topic, concerning the closure, does this work if it's an async/long running operation? It may depend only on the BLE timeout?

CoEpi/ble/Peripheral.swift
```diff
@@ -107,34 +117,26 @@ extension PeripheralImpl: CBPeripheralManagerDelegate {
...
-        let currentCENKey = CENKey.generateAndStoreCENKey()
-        let CENData: Data = CEN.generateCENData(CENKey: currentCENKey.cenKey!)
-        //*** Scenario 1: https://docs.google.com/document/d/1f65V3PI214-uYfZLUZtm55kdVwoazIMqGJrxcYNI4eg/edit#
-        // iOS - Central + iOS - Peripheral -- so commenting out addNewContact
-        //addNewContactEvent(with: identifier)
-        request.value = CENData
-        peripheral.respond(to: request, withResult: .success)
-        
-        addNewContactEvent(with: currentCENKey.cenKey!)
+        internalDelegate.onReceivedRequest(request: request.value, respondClosure: respondClosure)
```
i-schuetz: Does this work if `onReceivedRequest` is an async/long running operation? It may depend only on the BLE timeout?

CoEpi/domain/model/CENKey.swift
johnsonh: Should be consistent now

CoEpi/repo/CENKeyRepo.swift
johnsonh: I see... lets do it outside of this PR then

CoEpi/ble/Peripheral.swift
johnsonh: I think this function is the delegate called from BT - `func peripheralManager(_ peripheral: CBPeripheralManager, didReceiveRead request: CBATTRequest)` I used `Data` here because that's the type of `request.value`. I think on the outside we can convert the data to a CEN type

johnsonh: For removing all the `Contact` classes, do you mean all of these as well?
https://github.com/Co-Epi/app-ios/blob/develop/CoEpi/domain/model/Contact.swift
https://github.com/Co-Epi/app-ios/blob/develop/CoEpi/repo/realm/model/RealmContact.swift
https://github.com/Co-Epi/app-ios/blob/develop/CoEpi/repo/ContactRepo.swift
https://github.com/Co-Epi/app-ios/blob/develop/CoEpi/repo/realm/RealmContactRepo.swift

duskoo: Seems that for read requests, request.value is null:
`
/**
         2020-03-31 17:38:35.724037+0200 CoEpi[2303:2652797] [BLEPeripheral] Peripheral manager did receive read request:
         <CBATTRequest: 0x2803d81c0
             Central = <CBCentral: 0x2803d8480
                 identifier = 9E91D45A-64F4-1E48-E459-5C9068A36BBD,
                 MTU = 182
             >,
             Characteristic = <CBMutableCharacteristic: 0x2830b38e0
                 UUID = 2AC35B0B-00B5-4AF2-A50E-8412BCB94285,
                 Value = (null),
                 Properties = 0x2,
                 Permissions = 0x1,
                 Descriptors = (null),
                 SubscribedCentrals = (
                 )
             >,
             Offset = 0, Value = (null)
         >*/
`

CoEpi/ble/Peripheral.swift
johnsonh: The `DebugViewModel` does use it - `let peripheralContactSent = String(decoding: peripheralContactSent?.ourData ?? Data(), as: UTF8.self)`, because it displays when contacts are made. I realize its only for debugging, so we could remove this event and kill that functionality?

CoEpi/ble/Peripheral.swift
duskoo: We need to standardize on one Contact object. We have 3: Contact, ContactOld and BTContact.
If we are going to use BTContact, we should remove the other two.

i-schuetz: @johnsonh Yeah. Let's update with the API models (and the Android app, which orients with those). The `Contact` classes are mostly experimental.

- `Cen`
- `CenKey`
- `CenReport` (we don't need this for 0.1)

Additionally on Android I refined the `Cen` data type a bit. It seems that the timestamp is used only for _observed_ CENs. So I renamed the CEN+timestamp datatype in `ReceivedCen`. The CEN in `ReceivedCen` would be the actual `Cen` data type.

```kotlin
data class Cen(val bytes: ByteArray)
data class ReceivedCen(val cen: Cen, val timestamp: Int)
```

CoEpi/ble/Peripheral.swift
i-schuetz: My bad. Yes, we should leave it. But make it more explicitly for debugging `didReadCharacteristic`, `didReadCen` or similar.

duskoo: Current code is not decoding ourData properly, resulting in jumbled string.

CoEpi/ble/Peripheral.swift
johnsonh: Working on removing the others

johnsonh: Ok, I've removed `ContactOld`. I'd propose that we remove `Contact` related code in a separate PR, since this one is already huge and that seems like a discrete set of work.

johnsonh: So you're saying we should
- Update `Cen` to be `ReceivedCen`, like in Android
- Add API models?

I can do the former, but can we do the latter in a follow up PR?

johnsonh: Added a TODO section in the PR description

CoEpi/ble/Peripheral.swift
johnsonh: It depends, I mean saving to DB is an I/O operation ("long running"), would we want to `peripheralContactSent` to emit an event before persistence finished? Or if persistence errored out? I actually don't know what the desired behavior is,.

CoEpi/ble/Peripheral.swift
johnsonh: Removed `ContactOld`, I think we should remove `Contact` in a separate PR - added in the todo in the PR description

i-schuetz: I'll approve. You can decide when you want to do these changes.

CoEpi/ble/Peripheral.swift
i-schuetz: The actual question is for what are we persisting the generated CEN?

CoEpi/ble/Peripheral.swift
johnsonh: Renamed to `didReadCharacteristic`

johnsonh:
Merging into develop, to help us test and iterate faster. 
IMG_3A5EB0DE9040-1

IDs seem to be decoding correctly but we aren't sure

PR #47, opened 2020-04-02
Isolate CEN logic, components wiring

i-schuetz:


i-schuetz:
@rodneywitcher I put all the CEN logic in a file (unsurprisingly named `CenLogic`) and refactored a bit to make the flow clearer. Could you please go quickly through this file and confirm everything is correct?

CoEpi/domain/logic/CenLogic.swift
```diff
@@ -0,0 +1,67 @@
+import Foundation
+import CryptoSwift
+
+class CenLogic {
```
johnsonh, line 4: CenGenerator?

CoEpi/domain/logic/CenLogic.swift
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        //convert timestamp to [UInt8]
+        var tsAsUInt8Array: [UInt8] = []
+        [roundedTimestamp(ts: currentTs)].withUnsafeBytes {
+            tsAsUInt8Array.append(contentsOf: $0)
+        }
+
+        //encrypt tsAsUnit8Array using decodedCENKey... using AES
+        let encData = try! AES(key: decodedCENKeyAsUInt8Array, blockMode: ECB(), padding: .pkcs5).encrypt(tsAsUInt8Array)
```
johnsonh, line 44: Could reduce all of this to a computed var  on CENKey, or have a toData on CEN

duskoo: CenLogic could be part of CenKey class, but this also works.

CoEpi/domain/logic/CenLogic.swift
i-schuetz: It does more...

CoEpi/domain/logic/CenLogic.swift
i-schuetz: That could work, though what if we wanted to try out different generation algorithms? We wouldn't be able to swap those easily.

I'd perhaps consider splitting this class in `CenKeyGenerator` and `CenGenerator`?

PR #69, opened 2020-04-05
Matching calculations modifications and benchmarking

i-schuetz:
I'm not quite sure about the matching calculations, mostly just trying to understand.


CoEpi/domain/logic/CenMatcher.swift
```diff
@@ -19,26 +20,31 @@ class CenMatcherImpl: CenMatcher {
...
 
         var possibleCENs: [String] = []
         possibleCENs.reserveCapacity(max)
 
-        let CENKeyLifetimeInSeconds = 7*86400 // every 7 days a new key is generated
-
         for i in 0...max {
             let ts = maxTimestamp - Int64(CENLifetimeInSeconds * i)
-            let cen = cenLogic.generateCen(CENKey: key.cenKey)
+            let cen = cenLogic.generateCen(CENKey: key.cenKey, timestamp: ts)
```
duskoo, line 40: @i-schuetz My understanding is that CEN generation should only be done with rounded timestamps. This is currently not the case.

CoEpi/domain/logic/CenMatcher.swift
```diff
@@ -19,26 +20,31 @@ class CenMatcherImpl: CenMatcher {
...
-        let max = Int(7*24*(Double(60)/Double(CENLifetimeInSeconds)))
+        let max = Int(Double(7*24*60*60)/Double(CENLifetimeInSeconds))
 
         var possibleCENs: [String] = []
         possibleCENs.reserveCapacity(max)
 
-        let CENKeyLifetimeInSeconds = 7*86400 // every 7 days a new key is generated
-
         for i in 0...max {
             let ts = maxTimestamp - Int64(CENLifetimeInSeconds * i)
```
duskoo, line 39: @i-schuetz We should log and compare calculated timestamps between iOS and Android . Does Android also count from 1970?

CoEpi/domain/logic/CenMatcher.swift
i-schuetz: The rounding is done in `generateCen`

CoEpi/domain/logic/CenMatcher.swift
i-schuetz: Android has the same problem. And yes, it also uses unix epoch.

CoEpi/domain/logic/CenMatcher.swift
duskoo: Performance wise it would be better to do one initial rounding: maxTimestamp -> roundedMaxTimestamp and then only decrement that value by ( i * CENLifetimeInSeconds).

CoEpi/domain/logic/CenMatcher.swift
```diff
@@ -19,26 +20,31 @@ class CenMatcherImpl: CenMatcher {
 
     // Copied from Android implementation
     private func match(key: CENKey, maxTimestamp: Int64) -> [CEN] {
+
+        // Unclear why maxTimestamp is a parameter
+        let maxTimestamp = Date().coEpiTimestamp
```
duskoo, line 25: @i-schuetz 
Also not sure about the logic, but if we are to write unit tests for this function (and we definitely should), prviding maxTimestamp as input param is the way to go (parameter injection).

CoEpi/Dependencies.swift
```diff
@@ -65,14 +68,11 @@ class Dependencies {
     }
 
     private func registerWiring(container: DependencyContainer) {
-        container.register(.singleton) { ScannedCensHandler(coepiRepo: try container.resolve(),
+        container.register(.eagerSingleton) { ScannedCensHandler(coepiRepo: try container.resolve(),
```
duskoo, line 71: If eagerSingleton combined with container.bootstrap() instantiates ScannedCensHandler - great :)

CoEpi/domain/logic/CenMatcher.swift
i-schuetz: This is meant mostly as a reminder to clarify the issue. I'd have deleted the parameter otherwise. I want to confirm that it's always the current timestamp.

CoEpi/domain/logic/CenMatcher.swift
i-schuetz: Good point. For now I think it's better to keep the CEN logic "as is". The arithmetic to round the timestamp is unlikely to have a significant impact on performance. Let's come back to it when we've figure out how to address the actual problem.

CoEpi/Dependencies.swift
i-schuetz: It does now 🙂

i-schuetz: It does 🙂

PR #71, opened 2020-04-06
Onboarding screen shows only on first launch

ericlavigne:


CoEpi/domain/model/KeyValueStore.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+class KeyValueStore {
+    static let boolDefaults = ["showOnboarding": true]
```
i-schuetz, line 4: We don't need to cache the values. `UserDefaults` already does this.

CoEpi/domain/model/KeyValueStore.swift
```diff
@@ -0,0 +1,26 @@
...
+    
+    static func putBool(key: String, value: Bool) {
```
i-schuetz, line 6: I'd make an enum for the keys, to make it safer. E.g:

```swift
enum PreferenceKey {
    case seenOnboardidng
}
```
See the [Android implementation](https://github.com/Co-Epi/app-android/blob/develop/app/src/main/java/org/coepi/android/system/Preferences.kt)

(You can ignore `clear` for now)

CoEpi/domain/model/KeyValueStore.swift
```diff
@@ -0,0 +1,26 @@
...
+        let defaultResult = boolDefaults[key]
+        if defaultResult == nil {
+            print("Error: Storing bool for invalid key: " + key)
+        } else {
+            UserDefaults.standard.set(value, forKey: key)
+        }
+    }
+    
+    static func getBool(key: String) -> Bool {
```
i-schuetz, line 15: I forgot to comment, don't use `static` here. The idea is to allow this component to be mocked. We can't do this with `static`.

CoEpi/domain/model/KeyValueStore.swift
```diff
@@ -0,0 +1,26 @@
...
+        let defaultResult = boolDefaults[key]
+        if defaultResult == nil {
+            print("Error: Getting bool for invalid key: " + key)
+            return defaultResult! // Will throw an error
```
i-schuetz, line 19: Let's avoid force unwrap (`!`), unless it's really unavoidable. Here it's not (this code will disappear either way if you remove `boolDefaults`. Just sayin').

CoEpi/domain/model/KeyValueStore.swift
ericlavigne: KeyValueStore no longer specifies initial values for each of the keys. Instead, if a key's value has not yet been set, then nil will be returned.

CoEpi/domain/model/KeyValueStore.swift
ericlavigne: Keys are now an enum.

CoEpi/domain/model/KeyValueStore.swift
ericlavigne: No longer using static methods.

CoEpi/domain/model/KeyValueStore.swift
ericlavigne: Force unwrap is no longer used.

CoEpi/domain/model/KeyValueStore.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+enum KVSBoolKey: String {
```
i-schuetz, line 3: Just a minor thing, I'd not make the keys specific to the value types.

CoEpi/ui/RootWireframe.swift
```diff
@@ -22,11 +21,12 @@ class RootWireFrame {
         window.makeKeyAndVisible()
 
         self.homeViewController = homeViewController
-
-        if showOnboarding {
+        
+        if KeyValueStore().getBool(key: KVSBoolKey.seenOnboarding) != true {
```
i-schuetz, line 25: `KeyValueStore` should be an injectable component (this allows to change its implementation / mock it). So wire it in `Dependencies` and retrieve here with `container.resolve()` (or maybe pass it directly to the initializer, like the rest of components).

PR #76, opened 2020-04-07
Handle HTTP errors and show in alert

i-schuetz:
Now HTTP error status are forwarded correctly as errors and shown in the alert + possible API error message.


CoEpi/api/CoEpiApi.swift
```diff
@@ -33,36 +33,32 @@ class CoEpiApiImpl: CoEpiApi {
...
-                    emitter(.success(obj))
-                } catch let error {
-                    emitter(.error(ApiError.error(
-                        message: "Couldn't parse reponse: \(error), " +
-                            "data: \(String(describing: String(data: data, encoding: .utf8)))")))
+                if let apiError = response.apiError {
+                    emitter(.error(apiError))
+                } else {
+                    guard let data = response.data else {
+                        emitter(.error(ApiError.error(message: "Response didn't have data")))
```
duskoo, line 40: Can "no response data" be a valid response?

i-schuetz: Wouldn't expect it for a `get` request.

i-schuetz:
@duskoo it would be a nice to have. I personally would recommend using Charles though (or another web proxy), it's nicer to use and has more functionality.

i-schuetz:
@duskoo anyway add the interceptor if you want? I ended adding it to Android because Charles isn't working with the new endpoint (it works for iOS though).

PR #77, opened 2020-04-07
Use hex encoding

i-schuetz:
On hold. Unclear whether it should be Base64 or hex.


CoEpi/domain/logic/CenLogic.swift
```diff
@@ -61,7 +61,7 @@ class CenLogic {
             SecRandomCopyBytes(kSecRandomDefault, keyDataCount, mutableBytes)
         }
         if result == errSecSuccess {
-            return .success(keyData.base64EncodedString())
+            return .success(keyData.toHex())
```
duskoo, line 64: @i-schuetz 32 byte Data converted to Hex results in 64 chars.
It seems that v1 backend has 32 char column for cenKey...

https://github.com/Co-Epi/cen-server/blob/master/backend/cen.sql#L5

i-schuetz: Indeed. I tried with a 32 chars value and got the same error response. For now I'm waiting for clarification of whether we have to use Base64 or hex.

i-schuetz:
Yep, just happened to see it in the Android app too. Will change it 👍

PR #82, opened 2020-04-09
Update api to v1.api.coepi.org

i-schuetz:
I was able to do a successful match!

2 iOS devices, submit a report and it appears in the alerts of the other (ideally restart the app, since the fetch runs each hour).

Note that it takes a while for the alerts to be updated, since the matching is very slow (even with the 5-6 keys we have at the moment). @duskoo will add a progress indicator, so at least we see that something is happening.

Also added views to the debug screen to see my current key and CEN (the later can be deduced from the BLE data, but it's better to show it separately, especially if testing with more than 2 devices).


CoEpi/domain/logic/CenMatcher.swift
```diff
@@ -42,8 +44,12 @@ class CenMatcherImpl: CenMatcher {
             possibleCENs.append(cen.toHex()) // No fixed size array
         }
 
-        os_log("Generated results for key: %@, possible CENs: %@",
-               log: servicesLog, type: .debug, key.cenKey, "\(possibleCENs.count)")
+        os_log("Generated results for key: %@, possible CENs count: %@, CENs: %@", log: servicesLog, type: .debug, key.cenKey,
+               "\(possibleCENs.count)", "\(possibleCENs)")
```
duskoo, line 48: @i-schuetz 
There seems to be a problem with how the timestamps are rounded.
In 672 possible cens, during my test, there were only 2 unique values - which makes the whole processing questionable.
We should use CEN lifetime for rounding, not CENKey lifetime.
[possible-cens-problem.txt](https://github.com/Co-Epi/app-ios/files/4458549/possible-cens-problem.txt)

i-schuetz: That's a bug of the CEN implementation, yes. The rounding should use `CENLifetimeInSeconds`. I guess that it went undetected because we were using 120 seconds as `CENKeyLifetimeInSeconds` (which I changed to 1 week before uploading).

CoEpi/api/model/request/ApiParamsCenReport.swift
```diff
@@ -1,12 +1,15 @@
...
     let report: String
-    let cenKeys: [String]
+    let cenKeys: String
+    let reportTimestamp: Int64
 }
 
 extension ApiParamsCenReport {
     init(report: MyCenReport) {
-        self.init(report: report.report.report, cenKeys: report.keys)
+        self.init(reportID: report.id, report: report.report.report, cenKeys: report.keys.joined(separator: ","),
```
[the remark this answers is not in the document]
i-schuetz: I use heavily composition (in this case there's a `ReceivedCenReport` struct to differentiate from `CenReport`, which is reused in `MyCenReport` (which has the keys)) and they reference an internal `report` string. I prefer to call these fields simply `report`, making this chain inevitable.

PR #84, opened 2020-04-09
roundedTimestamp now takes roundingInterval parameter

duskoo:
When generating CENs, we round on CEN lifetime.
When generating CENKeys, we round on CENKey lifetime.

Also: Bluetooth reading enabled

### DIFF
--- a/CoEpi/ble/covidwatch/BluetoothController.swift
+++ b/CoEpi/ble/covidwatch/BluetoothController.swift
@@ -916,7 +916,7 @@ extension BluetoothController: CBPeripheralDelegate {
 
             // original code
             //let identifier = try UUID(dataRepresentation: value)
-            //self.logNewContactEvent(with: identifier)z
+            //self.logNewContactEvent(with: identifier)
             /****************************************************************/
 
         }
--- a/CoEpi/ble/covidwatch/BluetoothController.swift
+++ b/CoEpi/ble/covidwatch/BluetoothController.swift
@@ -526,8 +526,8 @@ extension BluetoothController: CBCentralManagerDelegate {
             else {
                 if (isConnectable) {
                     self.peripheralsToWriteContactEventIdentifierTo.insert(peripheral)
-                    // OR use reading. Both cases are handled.
-                    // self.peripheralsToReadContactEventIdentifierFrom.insert(peripheral)
+                    // OR use reading. Both cases are handled. - duskoo: Pretty sure we need to enable reading
+                    self.peripheralsToReadContactEventIdentifierFrom.insert(peripheral)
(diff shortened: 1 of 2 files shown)


CoEpi/ble/covidwatch/BluetoothController.swift
```diff
@@ -526,8 +526,8 @@ extension BluetoothController: CBCentralManagerDelegate {
             else {
                 if (isConnectable) {
                     self.peripheralsToWriteContactEventIdentifierTo.insert(peripheral)
-                    // OR use reading. Both cases are handled.
-                    // self.peripheralsToReadContactEventIdentifierFrom.insert(peripheral)
+                    // OR use reading. Both cases are handled. - duskoo: Pretty sure we need to enable reading
+                    self.peripheralsToReadContactEventIdentifierFrom.insert(peripheral)
```
i-schuetz, line 530: Hmm, wonder about the OR. These cases seem not related (write sends our CEN to other devices, while read gives us their CEN).

i-schuetz, line 530: This is a bit weird. According to But yeah at least according to https://docs.google.com/document/d/1f65V3PI214-uYfZLUZtm55kdVwoazIMqGJrxcYNI4eg/edit#heading=h.6q40wl39kcs8 at least, iOS to iOS should use characteristic read. And anyway if both devices write, they should see each other. But the timing here may not be right.

i-schuetz, line 530: This is a bit weird. According to https://docs.google.com/document/d/1f65V3PI214-uYfZLUZtm55kdVwoazIMqGJrxcYNI4eg/edit#heading=h.6q40wl39kcs8 at least, iOS to iOS should use characteristic read. And anyway if both devices write, they should see each other. But the timing here may not be right.

CoEpi/ble/covidwatch/BluetoothController.swift
```diff
@@ -916,7 +916,7 @@ extension BluetoothController: CBPeripheralDelegate {
 
             // original code
             //let identifier = try UUID(dataRepresentation: value)
-            //self.logNewContactEvent(with: identifier)
+            //self.logNewContactEvent(with: identifier)z
```
i-schuetz, line 919: Delete

CoEpi/ble/covidwatch/BluetoothController.swift
i-schuetz: (I assumed this fixes the issue I'm having with devices not seeing each other, but realized that it's not necessarily related. Will test later).

PR #90, opened 2020-04-10
Issue#89

Sam-gtri:
Implement share functionality

• Share button in home screen
• Sharing prompted after reporting symptoms


CoEpi/ui/home/HomeViewController.swift
```diff
@@ -32,13 +32,36 @@ class HomeViewController: UIViewController {
         viewModel.debugTapped()
     }
     
+    @objc func share(sender:UIView){
+       UIGraphicsBeginImageContext(view.frame.size)
+       view.layer.render(in: UIGraphicsGetCurrentContext()!)
```
i-schuetz, line 37: For what is this?

i-schuetz:
@Sam-gtri @danamlewis is submitting symptoms rare enough that showing the share dialog each time will not be annoying?

Sam-gtri:
Since this is a new app under very unique circumstances there are a few assumptions/hypothesis to test:

1) The type of people downloading this app will probably care about their well being and that of their family or others they know

2) People often ignore the first request to share something outright so continuing to prompt them until they do share will increase the growth of the app by insuring higher percentages of users share the app.

3) There will be a large number of people who only report symptoms once so prompting the share wont be annoying 

My goal is allow us to reach the most optimal growth through viral sharing so when we release the first version to testflight we can see from testers responses how they feel about the frequency

PR #96, opened 2020-04-11
Symptoms: concatenate strings, encode them in Base64 and send to API

duskoo:
IMG_6C6C83AD8365-1
Screen Shot 2020-04-11 at 5 07 48 PM

### DIFF
--- a/CoEpi/api/CoEpiApi.swift
+++ b/CoEpi/api/CoEpiApi.swift
@@ -5,7 +5,7 @@ import os.log
 
 protocol CoEpiApi {
 
-    func postCenReport(cenReport: MyCenReport) -> Completable
+    func postCenReport(myCenReport: MyCenReport) -> Completable
 
     func getCenKeys() -> Single<[String]>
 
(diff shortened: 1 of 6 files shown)


scottleibrand:
Merged and tested successfully between three iPhones.

CoEpi/api/model/request/ApiParamsCenReport.swift
```diff
@@ -9,7 +9,7 @@ struct ApiParamsCenReport: Encodable {
 
 extension ApiParamsCenReport {
     init(report: MyCenReport) {
-        self.init(reportID: report.id, report: report.report.report, cenKeys: report.keys.joined(separator: ","),
+        self.init(reportID: report.report.id, report: report.report.report.toBase64() ?? "Decoding error".toBase64()! , cenKeys: report.keys.joined(separator: ","),
```
i-schuetz, line 12: Let's examine why `toBase64()` returns an optional. Maybe the utf-8 decoding can be force unwrapped, see https://stackoverflow.com/a/46152738/930450

Otherwise we should look for a different way to handle this.

CoEpi/repo/SymptomRepo.swift
```diff
@@ -44,11 +44,10 @@ class SymptomRepoImpl: SymptomRepo {
...
 
 private extension Sequence where Iterator.Element == Symptom {
-
     func toCENReport() -> CenReport? {
-        // TODO (has not been specified yet)
-        "TODO symptoms -> CENReport".toBase64().map {
-            CenReport(id: "123", report: $0, timestamp: Date().coEpiTimestamp)
-        }
+        let separator = ", "
+        let stringReport : String  = self.reduce("") {$0 + $1.name + separator}
```
i-schuetz, line 49: You can use `map { $0.name }.joined(separator: ", ")`.  Then you don't need `trimmingCharacters`.
`self` not needed here.

CoEpi/api/CoEpiApi.swift
```diff
@@ -17,10 +17,10 @@ class CoEpiApiImpl: CoEpiApi {
 //    private let baseUrl = "https://q69c4m2myb.execute-api.us-west-2.amazonaws.com/v3/"
     private let baseUrl = "https://v1.api.coepi.org/"
 
-    func postCenReport(cenReport: MyCenReport) -> Completable {
-        os_log("Sending CEN report to API: %@", log: servicesLog, type: .debug, "\(cenReport)")
+    func postCenReport(myCenReport: MyCenReport) -> Completable {
```
i-schuetz, line 20: Personal preference: I prefer to keep parameter names simple when it's obvious from context. Ideally this would be called only `report`. But you can let it like this.

i-schuetz:
Left a few post-merge comments 🙂

PR #97, opened 2020-04-11
CEN matching: generate candidate CENs for each locally stored CEN

duskoo:
Fetch local CENs from Realm.
Fetch new keys from API.
For each local CEN:
   ts = cen.timestamp
   For each key:
      generate candidate CEN: f(key, rounded(ts))
      if localCen == candidatCen -> match found
...

For 2 local CENs matching time is ~4sec.
It should grow linearly with number of stored CENs.

### DIFF
--- a/CoEpi/domain/logic/CenMatcher.swift
+++ b/CoEpi/domain/logic/CenMatcher.swift
@@ -3,6 +3,7 @@ import os.log
 
 protocol CenMatcher {
     func hasMatches(key: CENKey, maxTimestamp: Int64) -> Bool
+    func matchLocalFirst(keys: [CENKey], maxTimestamp: Int64) -> [CENKey]
 }
 
 class CenMatcherImpl: CenMatcher {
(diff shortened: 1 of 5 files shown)


CoEpi/domain/logic/CenMatcher.swift
```diff
@@ -17,6 +18,39 @@ class CenMatcherImpl: CenMatcher {
...
+        
+        for localCen in localCens {
+            let mod = localCen.timestamp % Int64(CenLogic.CENLifetimeInSeconds)
+            let roundedLocalTimestamp = localCen.timestamp - mod
+            os_log("Local CEN: cen = [ %@ ], timestamp = [ %lld ], rounded timestamp = [ %lld ]", localCen.CEN, localCen.timestamp, roundedLocalTimestamp)
+            var i : Int = 0
+            for key in keys {
+                i+=1
+                let candidateCen = cenLogic.generateCen(CENKey: key.cenKey, timestamp: roundedLocalTimestamp)
+                let candidateCenHex = candidateCen.toHex()
```
i-schuetz, line 42: You could decode the stored CENs to data at the beginning, avoiding having to convert to hex here.

duskoo: @i-schuetz Ive opened new PR against develop: https://github.com/Co-Epi/app-ios/pull/103
Will look into the suggested changes (but not immediately)

PR #99, opened 2020-04-12
CR comments addressed

duskoo:


CoEpi/api/model/request/ApiParamsCenReport.swift
```diff
@@ -9,7 +9,8 @@ struct ApiParamsCenReport: Encodable {
 
 extension ApiParamsCenReport {
     init(report: MyCenReport) {
-        self.init(reportID: report.report.id, report: report.report.report.toBase64() ?? "Decoding error".toBase64()! , cenKeys: report.keys.joined(separator: ","),
+        //Utf-8 decoding can be force unwrapped, see https://stackoverflow.com/a/46152738/930450
+        self.init(reportID: report.report.id, report: report.report.report.toBase64()! , cenKeys: report.keys.joined(separator: ","),
```
i-schuetz, line 13: Thanks for addressing the comments! What I meant however was to put the force unwrap in the `toBase64` implementation, where we do the utf-8 decoding. This way it doesn't have to return an optional anymore.

duskoo: toBase64() is defined in CryptoSwift

i-schuetz: It is, but it's not the one we're using here.

duskoo: Right.
Well, its just a matter of where we put the exclamation mark then. 
:|

i-schuetz: Putting the exclamation mark here or in the `utf-8` unwrapping is different. In the `utf-8` unwrapping we're accepting exactly the `utf-8` force unwrapping as safe. While outside, we're applying it to the `toBase64()` function. The `utf-8` decoding is an implementation detail and it's not safe to make assumptions about it.

Besides we'd have to deal with optionals in each place where `toBase64()` is called.

duskoo: @i-schuetz 
> The utf-8 decoding is an implementation detail and it's not safe to make assumptions about it.
This is in line with current state of PR (accepting that toBase64() returns optional).
If you are confident there will be no issues, feel free to modify the PR and change the method to return String instead.

i-schuetz: > accepting that toBase64() returns optional 
🤔We're force unwrapping the optional, and the reason we're doing that, is because we accepted that the utf-8 decoding is safe (in general). Which means both that we're relying on an implementation detail and that we could force unwrap directly this decoding.

Alternatively, if we don't want to accept utf-8 as safe, we probably shouldn't force unwrap here either.

i-schuetz: Update: We actually don't have to force unwrap. It seems we can use this instead:
```
func toBase64() -> String {
    Data(utf8).base64EncodedString()
}
```

PR #100, opened 2020-04-12
Fix for Issue #102

duskoo:


CoEpi/ui/common/ConfirmationAlert.swift
```diff
@@ -10,16 +10,20 @@ struct ConfirmationAlert {
                                                 message: message,
                                                 preferredStyle: .alert)
 
-        alertController.addAction(UIAlertAction(title: noText, style: .default, handler: { action in
-            viewController.dismiss(animated: true, completion: {
-                noAction()
-            })
+        alertController.addAction(
+            UIAlertAction(title: noText, style: .default, handler: { action in
+                DispatchQueue.main.async {
```
i-schuetz, line 15: Hmm, why do we need this? I just investigated and apparently we don't even need to dismiss the alert? https://stackoverflow.com/questions/30853387/uialertaction-completion-block-not-called-ios/30853553

(I just tested on iOS 13 commenting out `viewController.dismiss(animated: true, completion: nil)` and it was dismissed. Could this maybe fix the issue on iOS 12?)

i-schuetz, line 15: I think the problem is that we shouldn't dismiss the view controller https://stackoverflow.com/questions/30853387/uialertaction-completion-block-not-called-ios/30853553

duskoo: The code in develop does not work on iOS 12. This is a bugfix. 
Ill check if not dismissing the vc solves the issue.

duskoo: @i-schuetz You are right: sharing works fine without dismissing VC explicitly.
Issue #102 requests removal of the "Share on submit", so I removed it.

i-schuetz: Ok! Can we leave the generic confirmation alert displayer? I'm all for removing not used code, but a confirmation alert is _very_ likely to be needed at some point.

i-schuetz: Though, thinking again, let's remove it just for the sake of consistency. Otherwise we may end keeping other things which are less likely to be needed.

PR #108, opened 2020-04-13
Multithreaded matching

duskoo:
Perform matching on multiple threads using concurrent dispatch queue and dispatch groups.
Performance improvement example: 
(iPhone 6 - 2 cores) Single threaded 17 sec -> Multithreaded  11sec


i-schuetz:
Very good. I was wondering whether there could be a threshold where the database `in` query becomes more performant, but realized that the number of stored CENs never will be too high (at most 10.000 for the last 3 weeks or so). The main factor will be the key count and the previous implementation is likely to just get worse, given all the work that it has to perform per key.

PR #112, opened 2020-04-15
Integrate TCN library as external dependency

i-schuetz:
Notes:
- Removed some entries from debug view ("peripheral did write", etc.). These informations are not provided by the external interface of the library. I believe nobody was looking at those anyway.
- The BLE service is now started in `BleAdapter` which is instantiated eagerly. This should be equivalent to starting it in the `AppDelegate`.
- Only using their BLE for now, not the TCN functionality.
- The library has iOS 13 as min deployment target. There seem to be plans to add support for iOS 12 but it's unclear when.

For now leaving the PR as draft while situation with iOS 12 is cleared.


i-schuetz:
Update: @danamlewis said it's ok to merge.

PR #118, opened 2020-04-18
Make timestamp type safe

i-schuetz:
This helps to ensure that we're actually using Unix time everywhere.

I considered not exposing the fact that the internal value is Unix time, but I think this would make the solution too complicated and possibly affect performance, as we'd not be able to operate externally with the timestamps.


i-schuetz:
Actually it probably makes more sense to call this class `UnixTime`. `CoEpiDate` would be meaningful if we wanted to change the internal representation, but this would break e.g. the database and it's not hiding it anyway.

CoEpi/domain/model/UnixTime.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+
+struct UnixTime: Codable {
+    let value: Int64
+
+    static func fromUnixTime(unixTime: Int64) -> UnixTime {
```
duskoo, line 6: Confusing: method creates UnixTime object from Int64.  Default initializer can be used instead.

i-schuetz: Right, forgot to update this.

CoEpi/domain/model/UnixTime.swift
```diff
@@ -0,0 +1,19 @@
...
+
+struct UnixTime: Codable {
+    let value: Int64
+
+    static func minTimestamp() -> UnixTime {
+        UnixTime(value: 0)
+    }
+
+    static func now() -> UnixTime {
+        UnixTime(value: Int64(Date().timeIntervalSince1970))
```
duskoo, line 11: @i-schuetz This might be a good point to switch to millisecond resolution that V4 backend is expecting.

i-schuetz: There has not been a final decision on the backend resolution yet.

Issue #116, opened 2020-04-18
Include previous time interval in the CEN candidate generation


i-schuetz:
Currently there can be a transition to the next rounded timestamp in the delay between the CEN generation by the sender and the storing by the receiver (which sets its timestamp to "now"). 

With the current algorithm, this should be fixed by generating a candidate CEN for the previous rounded time stamp too.

PR #119, opened 2020-04-18
Check matches for previous time interval (#116)

duskoo:


CoEpi/domain/logic/CenMatcher.swift
```diff
@@ -56,26 +56,39 @@ class CenMatcherImpl: CenMatcher {
...
-            let candidateCen = cenLogic.generateCen(CENKey: key.cenKey, timestamp: roundedLocalTimestamp)
-            let candidateCenHex = candidateCen.toHex()
-            os_log("%p %d. candidateCenHex: [%{public}@] based on key [%{public}@ %lld] \n", Thread.current, i, candidateCenHex, key.cenKey, key.timestamp)
-            if cen.CEN == candidateCenHex {
-                os_log("Match found for [ %{public}@ ]", candidateCenHex)
-                //Update matchedKeys on Main thread (preventing race conditions)
-                DispatchQueue.main.async{
-                    self.matchedKeys.append(key)
-                }
+            os_log("Key [ %d/%d ]", i, numOfKeys)
```
i-schuetz, line 65: Can we remove all these logs (including the ones in `matchKeyForTimestamp`)? This is too noisy for the repo. You can leave them commented, if you want.

duskoo: @i-schuetz Its a quick change. 
The idea with the abundant logs was for the advanced testers to be able to track the matching progress using the Console.app

PR #121, opened 2020-04-19
Backend v04 - Networking tests

duskoo:
Unit tests for forming and sending networking requests against V4 backend.
Tests are part of CoEpiNetworkingTests target and do not rely on app code - they can be executed independently. Only Alamofire lib is linked.


i-schuetz:
I'll review this as soon as we can confirm/clarify the backend spec (probably after 0.2 release).

Issue #95, opened 2020-04-11
Internationalization


ericlavigne:
Prepare for translation to other languages, starting with Spanish as an example.

https://developer.apple.com/internationalization/

ericlavigne:
Translated some of the .swift files using Localizable.strings files.

Branch: Internationalization

i-schuetz:
Looking good! Can you please also look into [SwiftGen](https://github.com/SwiftGen/SwiftGen)? It generates enums for the keys, which is safer and we get autocompletion.

ericlavigne:
Yes, I'll check out SwiftGen.

In the current approach, I think the .swift translations are easy enough to deal with due to explicit keys, but the .xib translations involve random keys that are unreadable. If SwiftGen has a better option for .xib translation, that would be very welcome.

i-schuetz:
I personally prefer to do the translations always in code. This way they in the same place and with SwiftGen it's very handy. When we transition to SwiftUI we'll need to have them in code anyway.

ericlavigne:
"I personally prefer to do the translations always in code."

That sounds like a good plan. Thanks.

ericlavigne:
All translations are now defined in Localizable.strings and applied in code. This already looks much nicer. I'll try SwiftGen next.

I applied symptom name translations within the "repo" code rather than in the UI. Are these names stored or sent over the network, so that using translated names might cause problems?

i-schuetz:
Currently they're just local. It's fine to translate directly in the repo. The idea is to remove this if/when the backend sends the symptoms (which would be already translated).

ericlavigne:
@i-schuetz The code for this issue is ready for review: https://github.com/Co-Epi/app-ios/pull/123

PR #123, opened 2020-04-21
Internationalization

ericlavigne:
Implements Internationalization, using Spanish as an example, as described in https://github.com/Co-Epi/app-ios/issues/95 .


CoEpi/ui/alerts/AlertsViewModel.swift
```diff
@@ -1,4 +1,5 @@
 import Dip
+import UIKit
```
i-schuetz, line 2: For what is `UIKit` needed?

ericlavigne: My original approach used a translation function from UIKit. Swiftgen replaced that function, so I will remove the UIKit import.

I'll have instructions added to README soon.

ericlavigne: Fixed both issues.

i-schuetz:
Consider moving the instructions to the wiki and linking from the readme. They're a bit too detailed for the readme.

ericlavigne:
> Consider moving the instructions to the wiki and linking from the readme. They're a bit too detailed for the readme.

I created a new Internationalization wiki page and changed the README to link to that page. Merging now.

PR #127, opened 2020-04-22
Disable button until a symptom has been checked

Sam-gtri:
Disables the submit button until a symptom is selected to prevent empty submissions


Sam-gtri:
@scottleibrand there already is a "none of the above" checkbox

CoEpi/ui/healthquiz/HealthQuizViewController.swift
```diff
@@ -37,7 +39,8 @@ class HealthQuizViewController: UIViewController, ErrorDisplayer {
 
     override func viewDidLoad() {
         questionList.register(cellClass: QuestionCell.self)
-
+        submitButton.isEnabled = false
```
i-schuetz, line 42: This should be reactive. Add an observable to the view model with enabled status. Derive enabled status from items observable (`selectedSymptoms` is not empty).

CoEpi/ui/healthquiz/HealthQuizViewController.swift
```diff
@@ -17,6 +18,7 @@ class HealthQuizViewController: UIViewController, ErrorDisplayer {
         
         dataSource.onChecked = { (question, idx) in
             viewModel.handleAnswer(question: question, idx: idx)
+            self.submitButton.isEnabled = true
```
i-schuetz, line 21: If the boxes are unchecked, the button would stay enabled.

scottleibrand:
Ok, as long as checking None of the Above still allows the report to be submitted, I withdraw my objection. Thanks.

PR #131, opened 2020-04-25
Add translations for 8 languages

ericlavigne:


i-schuetz:
Nice! Questions:
- How was this translated? I noticed 1-2 issues in the languages I know. Probably an automated translation service?
- Did the breaking of bold/non bold strings in the onboarding work well for all languages? This was accepted as a temporary (not localization friendly) solution. Ideally the translations should be the entire paragraphs and use markup to make parts bold. But if it's working we can let it like this (only for the onboarding though. If we have a similar use case in the future, we should implement it correctly).

ericlavigne:
"How was this translated? I noticed 1-2 issues in the languages I know. Probably an automated translation service?"

I used Google translate. The hard part was the beforeBold/bold/afterBold, because I needed to experiment with reverse translating chunks of the translated text until I found one that matched the part which should be bold.

"Did the breaking of bold/non bold strings in the onboarding work well for all languages? This was accepted as a temporary (not localization friendly) solution. Ideally the translations should be the entire paragraphs and use markup to make parts bold. But if it's working we can let it like this (only for the onboarding though. If we have a similar use case in the future, we should implement it correctly)."

As near as I could tell, this approach was adequate for all of these languages and for the small amount of text formatting that we have so far.

PR #136, opened 2020-04-26
Format alert time correctly

i-schuetz:


CoEpi/common/DateFormatters.swift
```diff
@@ -0,0 +1,17 @@
...
+
+class DateFormatters {
+    static let dateHoursMins: DateHoursMinsFormatter = DateHoursMinsFormatter()
+}
+
+final class DateHoursMinsFormatter: DateFormatter {
+
+    override init() {
+        super.init()
+        dateFormat = "dd.MM.yyyy HH:mm"
```
scottleibrand, line 11: This is only "correct" in Europe. We should use the locale date format instead, or something universal like yyyy-MM-dd HH:mm TZ.

PR #142, opened 2020-04-26
Update dateFormat

scottleibrand:
Use international standard date notation (YYYY-MM-DD), and include timezone.


i-schuetz:
Just curious, why do we show the users their timezone?

scottleibrand:
Because if you don't, I wonder if it's displaying in local time or UTC.

i-schuetz:
It's local time. I tested it when uploading the format.

scottleibrand:
Yeah, but it used to be UTC, so best to just show people it's PDT or CET so it's unambiguous.

scottleibrand:
image

scottleibrand:
Built locally and ran it on my phone, and it looks good to me.

PR #159, opened 2020-04-28
Issue117 fix

### LINKED ISSUE #117: Refactor home to use table (or collection) view
Before we add more changes to this screen, it should be refactored to show all the entries (including the debug data) in a table view. If needed, see debug view model for table view with mixed row types example.

astewart43-gtri:
Closes #117 

Simple refactor of home screen to table view.


i-schuetz:
This needs additional work. As discussed in the meeting, on hold until we decide whether we will support iOS 12.

i-schuetz:
It seems that we settled on supporting iOS 12 for now, but this PR needs changes. I think I said in the issue to orient with the debug view. We should be consistent.

i-schuetz:
Closing, as agreed in Slack. It requires more work and we need to focus on the other tasks.

PR #168, opened 2020-05-13
Rust library integration, proof of concept

i-schuetz:
See https://github.com/Co-Epi/app-backend-rust/pull/3

The idea is to integrate this PR in the current 0.4 unit tests (and possibly actual calls WIP) and remove from 0.1, so it's not broken for now. The integration in the 0.1 API is only meant as proof of concept.


i-schuetz:
The generated binary is currently ~177mb. Not sure if this is normal, have to investigate.

i-schuetz:
Update re: binary size, the final app size is 5mb - 14mb, which is about 1mb bigger than without the native framework.

PR #176, opened 2020-05-21
Wire Rust alert fetch, make core synchronous

i-schuetz:
This is part of the 0.4 protocol migration + the integration of the Rust lib. App is not functional (part of it still uses 0.1).


CoEpi/repo/AlertRepo.swift
```diff
@@ -1,35 +1,64 @@
...
+                    os_log("Inserted new alert: %{alert}@", log: servicesLog, type: .debug, "\(alert)")
+                }
+            }
+
+        case .failure(let error):
+            os_log("Error fetching alerts: %{public}@", log: servicesLog, type: .debug, "\(error)")
+            updateReportsStateSubject.accept(OperationState.failure(error: error))
+        }
+
+        updateReportsStateSubject.accept(.success(data: ()))
```
duskoo, line 60: @i-schuetz This line should be in case .success

CoEpi/repo/AlertRepo.swift
```diff
@@ -1,35 +1,64 @@
 import Foundation
 import RxCocoa
 import RxSwift
+import os.log
 
 protocol AlertRepo {
```
duskoo, line 6: "Alert" and "Report" are used as synonyms. We should standardize on one (I prefer report).

CoEpi/native/LibResult.swift
```diff
@@ -0,0 +1,50 @@
...
+extension Unmanaged where Instance == CFString {
+
+    func toLibResult<T>() -> LibResult<T> {
+        let resultValue: CFString = takeRetainedValue()
+        let resultString = resultValue as String
+
+        // TODO review safety of utf-8 force unwrap
+        let data = resultString.data(using: .utf8)!
+        let decoder = JSONDecoder()
+        return try! decoder.decode(LibResult<T>.self, from: data)
```
duskoo, line 48: Should we handle errors in decoding?

CoEpi/native/LibResult.swift
```diff
@@ -0,0 +1,50 @@
...
+typealias ArbitraryType = String
+
+extension LibResult {
+
+    func toResult() -> Result<T, ApiError> {
+        if isSuccess() {
+            if let data = data {
+                return .success(data)
+            } else {
+                return .failure(.error(message: "Unexpected: Library result success but no data: \(self)"))
```
duskoo, line 23: @i-schuetz  Empty response is a valid state. Lets just return .success(()) ?

CoEpi/repo/AlertRepo.swift
```diff
@@ -1,35 +1,64 @@
...
+        self.alertsFetcher = alertsFetcher
+        self.alertDao = alertDao
     }
 
     func removeAlert(alert: Alert) {
         cenReportsRepo.delete(report: alert.report)
     }
 
+    // TODO review thread safety with FFI/Rust: What happens if background task and foreground update
+    // TODO (e.g. pull to refresh) enter at the same time? Do we need to enqueue/lock?
```
duskoo, line 36: Background task could check what was the last update time.

CoEpi/repo/AlertRepo.swift
i-schuetz: I actually changed it in Android to be `Alert`, to differentiate better from outgoing reports.

CoEpi/native/LibResult.swift
i-schuetz: yup

CoEpi/native/LibResult.swift
i-schuetz: There's `toVoidResult()` for that.

CoEpi/repo/AlertRepo.swift
i-schuetz: That would make sense. This is more about possible race conditions (we're also doing early exit if there's already an operation in progress).

CoEpi/native/LibResult.swift
i-schuetz: I decided to return a `LibResult` with 502 (see note in code) instead of `Result<LibResult...>` to simplify the caller code.

CoEpi/repo/AlertRepo.swift
i-schuetz: I actually changed it in Android to be `Alert`, which is an incoming matched report. This differentiates it from non matching reports and outgoing reports.

PR #203, opened 2020-06-01
Fix homescreen buttons

### LINKED ISSUE #201: Implement home item backgrounds programmatically
Alternatively https://developer.apple.com/documentation/uikit/uiimage/1624102-resizableimagewithcapinsets

The images currently don't scale well.

mellowgeek:


mellowgeek:
@i-schuetz is there a resource file for the static values for corner radius and the shadows that I should put the values in instead of hard-coding it?

i-schuetz:
I'm fine with hardcoding for now. You also can create a file `Theme` or similar and put them as static constants there. There's no built-in solution for iOS like for Android.

PR #211, opened 2020-06-03
Refine symptoms flow

### LINKED ISSUE #190: Refine symptoms flow
Now that the navigation flow works, please go through all the report screens and complete anything that's not working or displaying incorrectly. If it's not minor and doesn't have an issue already, you can create one. @Co-Epi/ios

astewart43-gtri:
fixes some label formatting. 
closes #190


CoEpi/en.lproj/Localizable.strings
```diff
@@ -120,7 +120,7 @@
 "UX.cough.titleWet" = "Wet";
 "UX.cough.subtitleWet" = "\nIt feels like there is mucus \nin my throat and/or chest";
 "UX.cough.titleDry" = "Dry";
-"UX.cough.subtitleDry" = "\nMy throat feels 'tickly' or 'scratchy' without mucus";
+"UX.cough.subtitleDry" = "\nMy throat feels 'tickly' or \n'scratchy' without mucus";
```
i-schuetz, line 123: Please ensure that the PRs contain only the changes which are meant to be addressed. This is the only line that should have changed. Use an app like `Sourcetree` or Xcode's git diff to clean up the commits.

astewart43-gtri: My apologies. I noticed that in this and another PR after committing. What is the best way to partially revert a commit? should I simply remove the extra files and commit again?

astewart43-gtri: files have been removed

i-schuetz: Thanks. Next time you also can use e.g. [git reset](https://dev.to/konrad_126/git-reset---explain-49bl) in the command line

i-schuetz:
@astewart43-gtri This change closing #190 means that everything else in the report flow is working / displaying correctly or reported already?

astewart43-gtri:
actually, I noticed that I deleted a file by accident. that has been reverted.
Now it is correct. 
One small issue that I just noticed, the cough type page may have a visual error in Figma. Currently we have it implemented as shown in Figma, but it may be incorrect.
image

we are only able to select one of these options in the current design. should "Select all that apply" to "select one option"

i-schuetz:
Ok, can you bring that up in the ux channel?

PR #212, opened 2020-06-03
Input validation and correct keyboard type

### LINKED ISSUE #191: Input validation and correct keyboard type
Go through symptom flow inputs:
- Show correct keyboard (e.g. numeric for number inputs)
- Ensure there's no autocompletion etc. where not needed.
- Ask about handling of incorrect inputs (e.g. user copy pastes text in numeric keyboard) with UX. Since this seems a corner case, implement a simple solution (e.g. just clearing the input). 
- Min/max values and overflow handling.

astewart43-gtri:
closes #191


astewart43-gtri:
temperature input still needs some work. currently inserting a comma if the user inputs a number in the thousands, this causes some issues.

mellowgeek:
Temperature should never be in the thousands... might want to put a reasonable upper limit on that validation.

astewart43-gtri:
right, but you could have 101.4, meaning we need to support 5 digits. one possible solution would be to auto add a decimal after the third digit.

astewart43-gtri:
the issues with the temperature input are solved.

astewart43-gtri:
@i-schuetz currently this is accomplished only by restricting the input (length and keyboard type) and disabling paste. In my testing I am unable to break it. I can obviously add temperatures like 0.00. Should we also check the value when submit is tapped for more control?

ivnsch:
@astewart43-gtri Ideally yes, but if the UI makes it impossible to submit invalid values, let's not add additional validation for now. We can improve this in later versions.

CoEpi/common/NumberFormatters.swift
```diff
@@ -8,10 +8,10 @@ final class Number2MaxDecimalsFormatter: NumberFormatter {
 
     override init() {
         super.init()
-        numberStyle = .decimal
+        numberStyle = .none
         minimumFractionDigits = 0
-        maximumFractionDigits = 2
-        roundingMode = .halfUp
+        maximumFractionDigits = 0
+        roundingMode = .down
```
ivnsch, line 12: Why not decimals anymore? Also, if this is really necessary, it's probably enough to convert the value to integer and not use a formatter.

CoEpi/ui/symptomReport/FollowUp/Cough/CoughDaysViewController.swift
```diff
@@ -62,5 +62,21 @@ class CoughDaysViewController: UIViewController {
                 viewModel.onDaysChanged(daysStr: text ?? "")
             })
             .disposed(by: disposeBag)
+        
+        daysInput.rx.controlEvent(.editingChanged).subscribe(onNext: { [unowned self] in
```
ivnsch: Don't use `unowned`, please. It's a bit risky. Prefer `weak`.

CoEpi/ui/symptomReport/FollowUp/Fever/FeverDaysViewController.swift
```diff
@@ -62,5 +62,23 @@ class FeverDaysViewController: UIViewController {
                 viewModel.onDaysChanged(daysStr: text ?? "")
             })
             .disposed(by: disposeBag)
+        
+        
+        daysInput.rx.controlEvent(.editingChanged).subscribe(onNext: { [unowned self] in
```
ivnsch: Replace `unowned` with `weak`

CoEpi/ui/symptomReport/FollowUp/Fever/FeverTempViewController.swift
```diff
@@ -62,6 +62,12 @@ class FeverTempViewController: UIViewController {
                 viewModel.onTempChanged(tempStr: text ?? "")
             })
             .disposed(by: disposeBag)
+        
+        numberInput.rx.controlEvent(.editingChanged).subscribe(onNext: { [unowned self] in
```
ivnsch: `weak`

ivnsch (changes requested):

CoEpi/common/NumberFormatters.swift
astewart43-gtri: the style.decimal was responsible for the thousands comma formatting error (https://developer.apple.com/documentation/foundation/numberformatter/style/).
We cant just use int since we ned to support fractional numbers. This may be overkill, but I will leave that tip to you to decide. As it stands, I am unable to break the formatting and it is showing up correctly in the Xcode debug window.

astewart43-gtri:
> @astewart43-gtri Ideally yes, but if the UI makes it impossible to submit invalid values, let's not add additional validation for now. We can improve this in later versions.

That sounds good to me, I can open another issue regarding that. To clarify, it is currently possible to input unreasonable temperatures, such as 999. but it is not possible to force invalid characters.

ivnsch:
Ok, let's handle that for 0.4.

CoEpi/common/NumberFormatters.swift
ivnsch: Can you describe what exactly was the problem with the formatting? Can it be solved using [groupingSeparator](https://developer.apple.com/documentation/foundation/numberformatter/1412157-groupingseparator), maybe?

Either way it's wrong to change this specific formatter to this.

astewart43-gtri: when using "decimal" it auto corrected 1000 to 1,000.

ivnsch: You probably can remove the grouping separator using e.g. `groupingSeparator`. As said, this formatter shouldn't be used for this (name "Number2MaxDecimalsFormatter"). Otherwise, if you don't need formatting, don't use a formatter.

ivnsch (changes requested):
See formatting.

CoEpi/common/NumberFormatters.swift
astewart43-gtri: removing it is proving to be difficult. It seems that a formatter is needed without a drastic reworking. I will pick a more appropriate formatter.

ivnsch: Ok. Just create a new one.

astewart43-gtri:
new pull request opened to address these changes to avoid merge conflicts. Closing this one as duplicate.

PR #213, opened 2020-06-03
fixed breathless navigation

### LINKED ISSUE #210: breathless selection not in navigation.
When on the breathless page, selecting one of the options does not move you forward in the navigation. Skipping works fine and takes you to the next page in the navigation.

astewart43-gtri:
closes #210


CoEpi/ui/symptomReport/FollowUp/Breathless/BreathlessViewController.swift
```diff
@@ -69,20 +69,21 @@ class BreathlessViewController: UIViewController {
         
         tableView.rx.itemSelected
             .subscribe(onNext: {indexPath in
+                
                 if indexPath[1] == 0{
-                    print ("0")
+                    self.viewModel.onCauseSelected(cause: .leavingHouseOrDressing)
```
mellowgeek: Do you need to use self here?

astewart43-gtri: yes. without it you get "Reference to property 'viewModel' in closure requires explicit 'self.' to make capture semantics explicit"

CoEpi/ui/symptomReport/FollowUp/Breathless/BreathlessViewController.swift
```diff
@@ -69,20 +69,21 @@ class BreathlessViewController: UIViewController {
...
-                    print ("2")
+                    self.viewModel.onCauseSelected(cause: .groundOwnPace)
                 }
                 else if indexPath[1] == 3{
-                    print ("3")
+                    self.viewModel.onCauseSelected(cause: .hurryOrHill)
                 }
                 else if indexPath[1] == 4{
-                    print ("4")
+                    self.viewModel.onCauseSelected(cause: .exercise)
```
ivnsch: The array with the items should be in the view model. In another variable, in the view model, expose it to the view controller (using `Driver`, like in other view models). Here pass only the index: `viewModel.onCauseSelected(indexPath[1])`. In the view model, access the array using the index to get cause.

This is an acceptable solution. I personally prefer to not pass indices around but that would be more involved.

Also, you can declare `viewModel` in the capture list of the closure, to not have to use `self`: `.subscribe(onNext: { [viewModel] indexPath in`

ivnsch (changes requested):

CoEpi/ui/symptomReport/FollowUp/Breathless/BreathlessViewController.swift
ivnsch: General takeaways:
- Don't put data and logic in the UI (which includes view controllers). Thus view "model" - it's the backing model for the UI.
- If you're repeating blocks of code for each index/cell, like here, you probably want to access an array instead.
- Orient with existing code: The communication pattern between view models and view controllers is/should be the same everywhere.

astewart43-gtri: sounds good! will do!

CoEpi/ui/symptomReport/FollowUp/Breathless/BreathlessViewController.swift
```diff
@@ -43,15 +43,12 @@ class BreathlessViewController: UIViewController {
...
-        
         tableView.register(cellClass: UITableViewCell.self)
         tableView.tableFooterView = UIView()
         tableView.rowHeight = 70.0
         
         tableView.register(UITableViewCell.self, forCellReuseIdentifier: "Cell")
         
-        let images = ["house", "stop", "ground", "hill", "exercise"]
-        
         Observable.just([L10n.Ux.Breathless.p0, L10n.Ux.Breathless.p1, L10n.Ux.Breathless.p2, L10n.Ux.Breathless.p3, L10n.Ux.Breathless.p4])
```
ivnsch: This is also data and it should be in the view model.
Also, the translations, images and causes belong together. Orient with e.g. the alerts view model.

astewart43-gtri: yep, I moved that in the last commit.

ivnsch:
@astewart43-gtri Can this be reviewed (if yes, please don't forget to request review again in the right pane)?

astewart43-gtri:
No. I haven't had time to fix this yet.

ivnsch (approved):

PR #222, opened 2020-06-04
Update symptom report buttons

mellowgeek:


mellowgeek:
@i-schuetz made the suggested changes to the symptom report first, only issue is that there isn't any spacing between each item in the table view. Any suggestions? Other feedback also welcome.

i-schuetz:
@mellowgeek you should be able to solve that by adding top/bottom constraints with space to the button in the cells.

CoEpi/ui/styles/ButtonStyles.swift
```diff
@@ -0,0 +1,39 @@
...
+struct ButtonStyles{
+    
+    private init(){}
+    
+    static func applyStyleHomeCard(to: UIButton){
+        to.layer.cornerRadius = 15
+        applyStyleShadows(to: to)
+    }
+    
+    static func applyStyleSelectedButton(to:UIButton){
```
i-schuetz, line 13: Use local parameter name: `(to button: UIButon)`. `to:` is only for calling it.

CoEpi/ui/styles/ButtonStyles.swift
```diff
@@ -0,0 +1,39 @@
+import Foundation
+import UIKit
+
+struct ButtonStyles{
```
i-schuetz, line 4: This LGTM. It's not the fanciest styling system, but it works, until we start using SwiftUI and figure out best practices there.

PR #248, opened 2020-06-09
layout fixes and deleted unused assets.

### LINKED ISSUE #202: Improve layout on iPad / autolayout in general
The UI is pretty much broken on iPad. It's probably not mandatory to optimize for iPad yet, but it should be passable.

### LINKED ISSUE #244: Remove button background image
The asset `smallOvalButton` (and similar, if there are others) should be removed. Check that there are no more buttons referencing it. E.g. "What are exposure alerts" is using it.

mellowgeek:
Fixes height/width of "I don't know" and "Submit" buttons, touch backgrounds on the "How is your cough" view, updates the "what are exposure alerts" button on the alert screen, and removes unused background images.


CoEpi/ui/symptomReport/FollowUp/Cough/CoughHowViewController.swift
```diff
@@ -56,8 +56,28 @@ class CoughHowViewController: UIViewController {
         ButtonStyles.applyUnselected(to: worseButtonLabel)
         ButtonStyles.applyUnselected(to: sameButtonLabel)
         
+        betterButtonLabel.addTarget(self, action: #selector(onTouchDown(to:)), for:.touchDown)
```
ivnsch, line 59: Needs refactoring / better solution, but again, let's let it like this because of the SwiftUI migration.

ivnsch (approved):

CoEpi/ui/symptomReport/FollowUp/Cough/CoughHowViewController.swift
mellowgeek: I'll make an issue to track refactoring that.

PR #257, opened 2020-06-10
Temporary(app v0.3) filter for repeated TCNs

ivnsch:


astewart43-gtri:
Does this limit the number of TCN that can be processed at one time to 500?

ivnsch:
No, it blocks processing the TCN if it's contained in the latest 500 observed TCNs.

astewart43-gtri (approved):

PR #277, opened 2020-06-18
Adding SwiftLint and resolve warnings/errors

### LINKED ISSUE #218: Configure linter
And integrate in CI process. Linter errors should fail the build and not allow to merge PRs.

astewart43-gtri:
closes #218


ivnsch:
Small request in advance: Can we disable the identifiers min chars rule? This is IMO silly as it forbids to use e.g. `id`. Variables and method names with only one character should also be possible (we use this e.g. for the logs).

astewart43-gtri:
Yes, the latest push fixes that.
min length is set to 1 max is set to 60.

ivnsch:
Nice! Sure that length 1 isn't causing a warning though?

astewart43-gtri:
@i-schuetz This seems to be a ">=1" check. So no, it is working as desired.

.swiftlint.yml
```diff
@@ -0,0 +1,41 @@
+disabled_rules:
+- trailing_whitespace
```
ivnsch: Why is this disabled?

.swiftlint.yml
```diff
@@ -0,0 +1,41 @@
...
+- line_length
```
ivnsch: Why disabled? It's used below?

.swiftlint.yml
astewart43-gtri: My mistake, will correct.

.swiftlint.yml
astewart43-gtri: My mistake. I was testing and forgot to revert. Will correct.

.swiftlint.yml
```diff
@@ -0,0 +1,40 @@
+disabled_rules:
+- nesting
+- unused_closure_parameter
```
ivnsch: Why was this one added?

astewart43-gtri: nesting per our slack conversation (//Nesting Violation: Types should be nested at most 1 level deep (nesting))

unused_closure_parameter was a mistake. It needed to be aded while the auto corrector was running. it has been removed.

astewart43-gtri:
I made a best effort to go through the code and clean up inconsistent formatting. I likely missed a lot, but I think this is a good starting point.

ivnsch:
Meaning we have a lot of new warnings now? Can you tackle those in a new PR?

astewart43-gtri:
most of the existing warnings are related to TODO comments, force_cast, and force_try.
yes, everything can be addressed in a separate PR, assuming the formatting is what you want it to be.
like the function names that were too long, separating them into multiple lines in a way that makes sense.

ivnsch (approved):

duskoo:
@astewart43-gtri Current develop (c36a5bd) is not buildable due to too many swiftlint errors. Any suggestions?

astewart43-gtri:
it seems to build fine for me. 
However, if you need to ignore an error, you can add it under "disabled_rules:" heading in .swiftlint.yml

PR #280, opened 2020-06-19
swiftLint install in readme

astewart43-gtri:


duskoo (approved):

haleybear:
I installed swiftlint and have 3 build errors now regarding:
Swift Compiler Error Group:
.../app-ios/CoEpi/native/NativeCore.swift:79:27: Use of undeclared type 'CoreLogMessage'
.../app-ios/CoEpi/native/NativeCore.swift:65:34: Use of unresolved identifier 'register_log_callback'
.../app-ios/CoEpi/native/NativeCore.swift:73:13: Use of unresolved identifier 'CoreLogLevel'
Should I roll back my version to an earlier state?

astewart43-gtri:
@haleybear That is likely not due to swiftlint, but rather needing to rerun "carthage update"

PR #299, opened 2020-06-30
Upgrade core to 0.1.22

ivnsch:
This release doesn't have differences to 0.1.21 for iOS.


haleybear:
@i-schuetz I upgraded to this and I cannot run the application - the build succeeds but I'm getting an error still when trying to run the app in the emulator. I re-ran carthage bootstrap prior, but is there something else that I'm missing?

image

ivnsch:
This seems like Xcode weirdness, can you try to clean, clear contents in the simulator, restart Xcode etc? See also https://developer.apple.com/forums/thread/120282

Issue #250, opened 2020-06-09
Separate breathlessness images from background


ivnsch:
See https://github.com/Co-Epi/app-android/pull/152#discussion_r422685840

ivnsch:
Any updates? For now it would be enough to just separate the assets and show one image (icon) on top of the other (line background). I can do the rest of the refactoring later.

Sam-gtri:
@i-schuetz Ill work on this today!

PR #301, opened 2020-06-30
Separate Breathless images from background

Sam-gtri:
#250


ivnsch:
I'm a bit confused by the code: I only see that setting the image was removed?

Sam-gtri:
@i-schuetz I must have misread the android repo - I thought that the conclusion reached was that for the time being the idea was to remove the images so that more text options could be more easily added without having to generate related images.

PR #310, opened 2020-07-16
Integrate distance and duration

### LINKED ISSUE #311: Exposure alerts
**Individual exposure alerts**
1. Timeline of repeated interactions previously logged
2. Distance from you on exposure alert
3. Menu button in header to delete

Screen Shot 2020-07-21 at 6 08 03 PM

**Exposure alert screen** 

1. Red dots to indicate new alert - will disappear once clicked
2. Preview of first symptoms will show (similar to a text)
3. Blue boxed indicator to represent if an interaction is repeated (the user has already interacted with the same  person previously)
4. Definition of exposure alerts under title

Screen Shot 2020-07-21 at 7 48 37 PM

ivnsch:
PR 1 (from estimated 3) for https://github.com/Co-Epi/app-ios/issues/311

Screen Shot 2020-07-23 at 14 57 33

I didn't get an official OK for SwiftUI, so I'll smuggle it in ;P
As said, if we end supporting iOS 12 the screen can be ported very quickly to .xib
For now only for the details screen.


CoEpi/ui/alerts/AlertDetailsViewModel.swift
```diff
@@ -1,35 +1,61 @@
...
-
+    
     func formatReportTime(date: Date) -> String {
         let monthDay = DateFormatters.monthDay.string(from: date)
-
+        
         // For the weirdest reason, this formatter returns an optional, though it has exactly the same interface as
         // the month day formatter. Seems to be a compiler bug.
         // TODO revisit, maybe in future Xcode versions.
         let time = DateFormatters.hoursMins.string(for: date)!
```
duskoo, line 44: @i-schuetz Wrong method used. Should be:
```
string(from date: Date) -> String
```

PR #319, opened 2020-07-24
Move alerts to core, remove Realm

ivnsch:
Note: requires a new core release with https://github.com/Co-Epi/app-backend-rust/pull/130


CoEpi/extensions/Array+Convenience.swift
```diff
@@ -11,4 +11,11 @@ extension Array where Element: Equatable {
         }
         return uniqueValues
     }
+
+    func deleteFirst(element: Element) -> [Element] {
+        guard let index = firstIndex(of: element) else { return self }
+        var mutSelf = self
```
duskoo, line 17: @i-schuetz This reimplements copy-on-write logic that Swift arrays already have: https://developer.apple.com/documentation/swift/array
Seems unnecessary at first glance. Is it needed?

CoEpi/ui/alerts/AlertDetailsViewModel.swift
```diff
@@ -34,15 +34,15 @@ private extension Alert {
...
-    
+
     func formatReportTime(date: Date) -> String {
         let monthDay = DateFormatters.monthDay.string(from: date)
-        
+
         // For the weirdest reason, this formatter returns an optional, though it has exactly the same interface as
         // the month day formatter. Seems to be a compiler bug.
         // TODO revisit, maybe in future Xcode versions.
         let time = DateFormatters.hoursMins.string(for: date)!
```
duskoo:
```
let time = DateFormatters.hoursMins.string(from: date)
```

duskoo (approved):

CoEpi/ui/alerts/AlertDetailsViewModel.swift
ivnsch: Lol, I must have been tired.

CoEpi/extensions/Array+Convenience.swift
i-schuetz: You mean that it can be implemented like this:

```swift
mutating func deleteFirst(element: Element) -> [Element] {
    guard let index = firstIndex(of: element) else { return self }
    remove(at: index)
    return self
}
```

And called like this?

```
private func removeAlertLocally(alert: Alert) {
    let alertsState = alertsStateSubject.value
    switch alertsState {
    case .success(var alerts): // <-- var instead of let
        let newAlerts = alerts.deleteFirst(element: alert)
        alertsStateSubject.accept(.success(data: newAlerts))
    default: break
    }
}
```
This works, but I personally prefer an immutable api (not requiring to use `var`). But open to discuss.

ivnsch: Not sure how exactly it relates to copy on write. We could implement a mutating version, like this:

```swift
mutating func deleteFirst(element: Element) {
    guard let index = firstIndex(of: element) else { return }
    remove(at: index)
}
```

And call it like e.g. this:

```swift
private func removeAlertLocally(alert: Alert) {
    let alertsState = alertsStateSubject.value
    switch alertsState {
    case .success(var alerts):
        alerts.deleteFirst(element: alert)
        alertsStateSubject.accept(.success(data: alerts))
    default: break
    }
}
```
But I personally prefer an immutable api. Open to discuss.

duskoo: > Not sure how exactly it relates to copy on write. We could implement a mutating version, like this:
> ...
> But I personally prefer an immutable api. Open to discuss.

Yes, mutating version would be my first pick, but consistency is also important, so lets go with immutable approach.

PR #366, opened 2020-09-19
Multiple reminder notifications

duskoo:
On app start schedule n (n=14) daily reminder notifications.
On report submission, cancel the reminder notification for the current day.
Settings toggle:
off -> cancel all pending notifications
on -> schedule n daily reminder notifications

### DIFF
--- a/CoEpi/AppDelegate.swift
+++ b/CoEpi/AppDelegate.swift
@@ -4,6 +4,7 @@ import CoreData
 import Dip
 import UIKit
 
+
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
     let container: DependencyContainer = Dependencies().createContainer()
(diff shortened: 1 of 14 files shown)


CoEpi/AppDelegate.swift
```diff
@@ -23,6 +24,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         let bgManager: BackgroundTasksManager = try! container.resolve()
         bgManager.onApplicationFinishLaunching()
 
+        let notificationShower: NotificationShower = try! container.resolve()
+        notificationShower.scheduleReminderNotificationsForNext(days: 14)
```
duskoo, line 28: Number of days for which to schedule reminder notifications (in advance). 
Note: iOS limits number of pending local notifications to 64.

SymptomFlowManager.swift
```diff
@@ -111,6 +114,8 @@ class SymptomFlowManager {
     func clear() {
```
duskoo, line 114: clear is called as the last step of handling successful report submission

scottleibrand:
Nice. So as long as someone taps one of those 14 or otherwise opens the app, they’ll keep getting notified, and if they completely ignore it for 2 weeks it’ll leave them alone? I like it.

PR #372, opened 2020-10-17
AlertDetailsViewController onAppear: perform markLinkedAlertsAsRead

### LINKED ISSUE #371: Mark all Repeated interactions as read on reading one of them
When clicking a Repeated interaction exposure alert, it currently marks only that single interaction as read and clears the red Unread indicator. This often leaves dozens of others that still have to be cleared to stop getting notifications on every refresh.

We should instead mark as read (and clear the red Unread indicator) on all Repeated interactions whenever the user clicks on one of them.

Equivalent to https://github.com/Co-Epi/app-android/issues/353 on Android

duskoo:
#371 

Looping through linkedAlerts on the background thread and setting each as read individually.


CoEpi/ui/alerts/AlertDetailsViewModel.swift
```diff
@@ -58,6 +59,22 @@ class AlertDetailsViewModel: ObservableObject {
     func reportProblemTapped() {
         email.openEmail(address: "TODO@TODO.TODO", subject: "TODO")
```
duskoo: @danamlewis Real email address missing
Should I use: address: "CoEpi@OpenAPS.org", subject: "Problem with CoEpi" ?

scottleibrand: Yes please.

scottleibrand (approved):

CoEpi/ui/alerts/AlertDetailsViewModel.swift
danamlewis: fixed. thx for flagging.

